Issue #15, opened 2026-04-24
BUILD-002: Integrate spike prototypes into a single game/ Bazel workspace


cgruber:
Ticket: `thoughts/shared/tickets/BUILD-002-integrate-spikes-into-bazel.md`

Merge the SPIKE-001 game prototype (TypeScript/Vite/D3/Zustand) and SPIKE-002 Bazel PoC (rules_rust + rules_ts + Rust→WASM) into a single unified source root under `game/` with a single `bazel build //...` that produces the game.

See ticket file for full acceptance criteria, background, and definition of done.

cgruber:
Complete in sprint 0

PR #16, opened 2026-04-24
build(002): integrate spikes into unified game/ Bazel workspace

cgruber:
## Prerequisites
- [ ] N/A — no parent PR

## Summary
- Creates `game/` as a standalone Bazel workspace integrating spike/001 (TypeScript+D3+Zustand game) and spike/002 (Rust+WASM build PoC) into a single `bazel build //...` target
- Implements all BUILD-002 acceptance criteria: build passes, tests pass, `bazel run //:serve` serves the game at localhost:8080 with WASM diagnostic
- Key non-obvious wiring: esbuild Bazel 9.x patch, zustand/vanilla subpath to avoid React, no-modules WASM target with async HTML init sequence

## Validation performed
- [x] `bazel build //...` passes (6 actions, all cached after first run)
- [x] `bazel test //...` passes — rust:calc_test and web:app_typecheck_lib_typecheck_test both PASSED
- [x] `bazel run //:serve` starts HTTP server; HTML, bundle.js, and wasm_calc_bindgen.js all served correctly (verified via curl)

- Local dev only — no deployed services

- N/A — no deployment; local dev workspace only

## Tickets addressed
- see #15 (BUILD-002)

## Rollback
- Revert by dropping the `game/` directory; no other files affected


game/web/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+        "//:node_modules/d3",
+        "//:node_modules/@types/d3",
+        "//:node_modules/zustand",
+        "//:node_modules/zundo",
+    ],
+)
+
+# Alias used by bazel test — a no-op build that confirms type checking passes.
+alias(
+    name = "app_typecheck_test",
```
cgruber, line 25: `alias(name = "app_typecheck_test", actual = ":app_typecheck_lib")` does not produce a test target. An `alias` pointing to a `ts_project` (a non-test rule) will be silently ignored or fail under `bazel test //...` — TypeScript type-checking will **not** run as part of the test suite.

The BUILD-NOTES.md claim "bazel test //... — run all tests (Rust unit + TypeScript type-check)" is therefore inaccurate: only `//rust:calc_test` actually runs.

**Fix options:**
1. Use `build_test` from `@bazel_skylib//rules:build_test.bzl`:
   ```
   load("@bazel_skylib//rules:build_test.bzl", "build_test")
   build_test(
       name = "app_typecheck_test",
       targets = [":app_typecheck_lib"],
   )
   ```
2. Or use a `sh_test` that invokes a no-op script but declares `:app_typecheck_lib` in `data` (forces the build).

`build_test` is already available via `bazel_dep(name = "bazel_skylib", ...)` in MODULE.bazel, so no new dependency is needed.

cgruber:
Reviewed game/ Bazel workspace integration (category D — all new files, no existing files touched).

**Scope:** Clean. All 25 files are within `game/`. No scope creep.

**One issue found** (inline comment on `game/web/BUILD.bazel:24`): the `alias(name = "app_typecheck_test", ...)` pointing to a `ts_project` is not a test rule and will not run under `bazel test //...`. TypeScript type-checking is not actually exercised by the test suite. Fix: use `build_test` from `@bazel_skylib//rules:build_test.bzl` (no new dependency — bazel_skylib is already declared).

**Everything else is solid:**
- MODULE.bazel / bzlmod wiring is correct; esbuild Bazel 9.x patch is minimal and precise
- `rules_nodejs` 6.7.4 override to fix CcInfo removal is correctly motivated
- `zustand/vanilla` subpath + `react`/`react-dom` as esbuild externals is the right approach for a no-React bundle
- Rust code is clean; `fn main() {}` in `wasm_entry.rs` is correctly required for `rust_binary`; `#![deny(warnings)]` is appropriate
- TypeScript is strict (`noUncheckedIndexedAccess`, `exactOptionalPropertyTypes`, `strict`); null guards and D3 data-join pattern are correct
- `innerHTML` in the results renderers is safe — all content is internally generated numbers/constants with no user input
- Hermeticity claim in BUILD-NOTES.md is accurate with the documented `python3 -m http.server` caveat
- `pnpm-lock.yaml` checked in; `onlyBuiltDependencies: []` is correctly set for pnpm v9

Fix the `app_typecheck_test` alias and this is ready to merge.

game/web/BUILD.bazel
cgruber: Fixed. Replaced the `alias` with a `build_test` from `@bazel_skylib//rules:build_test.bzl`. The load statement was added at the top and the rule now reads:

```
build_test(
    name = "app_typecheck_test",
    targets = [":app_typecheck_lib"],
)
```

Verified: `bazel test //web/...` now shows both `//web:app_typecheck_lib_typecheck_test` and `//web:app_typecheck_test` PASSED.

PR #12, opened 2026-04-24
SPIKE-002: Bazel 9.1+bzlmod TypeScript+Rust/WASM build PoC

cgruber:
## Prerequisites
- [x] Parent PR merged: spike/001-game-poc (SPIKE-001) — merged as #11

## Summary
- Implements SPIKE-002: Bazel 9.1.0 + bzlmod + rules_rust 0.70.0 + aspect_rules_ts 3.8.8 build PoC in `spike/002-build-poc/`
- Proves TypeScript + Rust→WASM can be built hermetically from a single `bazel build //...`; `bazel test //...` passes 2 test targets
- Resolves both SPIKE-001 and SPIKE-002 in TICKETS.md

## Validation performed
- [ ] N/A — not an infra/k8s change
- `bazel build //...` passes from `spike/002-build-poc/` (all artifacts built hermetically)
- `bazel test //...` passes — `//rust:calc_test` (3 Rust unit tests) + `//web:app_typecheck_test`
- `bazel run //:serve` assembles outputs and serves on port 8080; TypeScript calls wasm_bindgen.add(2,3) and displays 5
- SPIKE-REPORT.md documents versions, hermeticity gaps, friction classification, build times, and go recommendation

## Affected machines / services
- None — spike source only; no deployment

## Deployment steps (POST-MERGE)
- None

## Tickets addressed
- see SPIKE-002 (thoughts/shared/tickets/SPIKE-002-harmonized-bazel-build-poc.md)
- TICKETS.md: marks SPIKE-001 and SPIKE-002 Resolved

## Rollback
- N/A — spike source; no production impact

PR #11, opened 2026-04-24
SPIKE-001: TypeScript+Vite+D3+Zustand game prototype

cgruber:
## Prerequisites
- [ ] N/A — base PR (targets main)

## Summary
- Implements SPIKE-001: TypeScript + Vite + D3.js + Zustand game prototype in `spike/001-game-poc/`
- Proves the browser-side game tech stack: hex-grid map generation, SVG/D3 rendering, Zustand state management, election simulation, and boundary rendering
- All acceptance criteria met; SPIKE-REPORT.md written; go recommendation for this stack

## Validation performed
- [ ] N/A — no Bazel or infra targets
- All game logic tests pass (`npm test`) — hex grid, election sim, boundary edges
- `npm run dev` served locally; hex map renders, election simulation runs, district coloring works
- SPIKE-REPORT.md documents friction, build times, and go/no-go recommendation

## Affected machines / services
- None — spike source only; no deployment

## Deployment steps (POST-MERGE)
- None

## Tickets addressed
- see SPIKE-001 (thoughts/shared/tickets/SPIKE-001-game-tech-stack-poc.md)

## Rollback
- N/A — spike source; no production impact

PR #9, opened 2026-04-23
Add kotlin tools to critique/response agent prompts in pr-review-cycle

cgruber:
## Summary

- Rewrites `thoughts/shared/workflows/pr-review-cycle.md` to use kotlin scripts throughout instead of raw `gh api` calls
- Adds compressed companion `pr-review-cycle.compressed.md` (was missing)
- Adds AGENT-002 ticket tracking the same fix needed in the infra repo's internal workflow doc
- Updates TICKETS.md with AGENT-002

Key changes in the workflow doc:
- Step 0: adds `gh-pr-diff-summary` classification step (was missing entirely)
- Step 2: switches to `gh-pr-create.main.kts` (validates body sections before creating)
- Step 3 critique prompt: replaces "Use `gh api` to post comments" with explicit kotlin script preamble listing `gh-pr-comment.main.kts` and `gh-pr-threads.main.kts`
- Step 4 response prompt: same tools preamble
- Step 7: replaces raw GraphQL thread resolution with `gh-pr-threads.main.kts`
- Step 8: removes `--auto` flag; adds `gh-pr-status.main.kts` pre-merge check
- Fixes all `geekinasuit/polyglot` repo references to `cgruber/redistricting-sim`

## Validation performed

- [ ] N/A — workflow documentation only

## Tickets addressed

see AGENT-002

PR #4, opened 2026-04-23
Remove Kotlin/Bazel; add SPIKE-001 and SPIKE-002 tickets

cgruber:
## Summary

- Removes the Kotlin gRPC service (`kotlin/`, `protobuf/`, `db/`, `util/`) — game will be a TypeScript browser app
- Removes all Bazel build infrastructure (`MODULE.bazel`, `BUILD.bazel`, `.bazelversion`, `.bazelrc`) — build system being re-evaluated via SPIKE-002
- Updates `.gitignore` for TypeScript/Rust/WASM toolchains
- Strips `.buildkite/pipeline.yml` to a placeholder (stack being decided)
- Rewrites `AGENTS.md` and `AGENTS.compressed.md` for post-Kotlin state; adds `SPIKE-NNN` ticket category; documents spike directory isolation rules
- Rewrites `README.md` to reflect current working direction
- Adds `SPIKE-001` ticket: TypeScript + SVG/D3 + Zustand game proof-of-concept in `spike/001-game-poc/` (independent source root, includes TypeScript quality standards for agents)
- Adds `SPIKE-002` ticket: harmonized Bazel build proof-of-concept in `spike/002-build-poc/` (independent source root, includes Bazel + Rust quality standards for agents)
- Updates `TICKETS.md`: closes superseded KT/BUILD/CI tickets; adds SPIKE category; references polyglot exemplar for SPIKE-002

## Test plan

- [x] `kotlin/`, `protobuf/`, `db/`, `util/` directories removed
- [x] `MODULE.bazel`, `MODULE.bazel.lock`, `BUILD.bazel`, `.bazelversion`, `.bazelrc` removed
- [x] `AGENTS.md` and `AGENTS.compressed.md` contain no references to Kotlin, ktfmt, or Bazel build conventions
- [x] SPIKE-001 ticket documents TypeScript quality standards (strict mode, Biome, module structure, D3/Zustand idioms)
- [x] SPIKE-002 ticket documents Bazel + Rust quality standards (bzlmod, rules_ts, rules_rust, wasm-bindgen discipline)
- [x] Both spike tickets specify `spike/NNN-*/` as independent source root with jj workspace isolation instructions
- [x] `TICKETS.md` superseded tickets moved to Resolved with rationale

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24, opened 2026-04-25
docs: scenario data format spec

cgruber:
## Prerequisites
- [ ] N/A

## Summary

Scenario data format spec — the gate before simulation implementation (per the election simulation ADR).

Defines the canonical structure for a redistricting scenario file, covering:
- Top-level scenario shape: region, geometry, parties, districts, precincts
- Demographic groups with optional dimensional schema and completeness constraint
- Voter eligibility derived from `eligibility_rules` (not on groups)
- Demographic events (`turnout_shift`, `vote_share_shift`, `population_shift`) firing at evaluation time
- Scenario rules (population tolerance, contiguity, compactness threshold)
- Success criteria discriminated union (`seat_count`, `majority_minority`, `efficiency_gap`, `mean_median`, `compactness`, `safe_seats`, `competitive_seats`, `population_balance`)
- Narrative content (character framing, intro slides, objective)
- State context (pre-computed other-region results; parsed but renderer-optional in v1)
- Geometry: `hex_axial` (adjacency derived from q,r) vs `custom` (explicit neighbor lists)
- Unassigned precinct handling: auto-fill editable precincts to `districts[0]`; context precincts require `initial_district_id`
- 8 open questions documented

## Validation performed
- [ ] N/A — specification document only; no code changes

- None

- N/A

## Tickets addressed
- N/A (this spec is itself the gate referenced in the election simulation ADR)

## Rollback
- N/A — documentation only

### DIFF
--- a/thoughts/shared/decisions/2026-04-24-scenario-data-format.md
+++ b/thoughts/shared/decisions/2026-04-24-scenario-data-format.md
@@ -1,0 +1,597 @@
+---
+date: 2026-04-24
+status: draft — awaiting review
+---
+
+# Spec: Scenario Data Format
+
+## Status
+
+Draft. This document is the gate before simulation implementation — the evaluation
+engine must not be built until this spec is reviewed and accepted.
+
+## Purpose
+
+Defines the canonical data format for a redistricting scenario. The format serves:
+
+- Pre-built scenario authoring (hand-authored JSON/YAML)
+- The simulation and evaluation engine (reads this directly)
+- Future player-authored scenarios and community sharing (post-v1)
+- Scenario persistence / export
+
+All data is self-contained in one document per scenario. No external lookups at
+runtime. A scenario file is deterministic: given the same file and the same player
+map, the simulation always produces the same result.
+
+## Conventions
+
+- **Format**: JSON canonical; YAML acceptable for hand-authoring (must round-trip
+  cleanly to JSON). All field names: `snake_case`.
+- **Identifiers**: opaque strings. `ScenarioId`, `PrecinctId`, `DistrictId`,
+  `PartyId`, `GroupId`, `RegionId` are all `string`. Identifiers must be unique
+  within their scope (e.g., `PrecinctId` unique within scenario).
+- **Floats**: all fractional values are `[0.0, 1.0]` unless stated otherwise.
+  Populations are non-negative integers.
+- **Format version**: `"format_version": "1"` in every scenario file. Parser
+  must reject unknown versions.
+
+---
+
+## Type Definitions
+
+### Top-level: `Scenario`
+
+```typescript
+interface Scenario {
+  format_version: "1"
+  id:             ScenarioId
+  title:          string
+  election_type:  ElectionType        // "congressional" | "state_senate" | "state_house"
+
+  region:         RegionSpec
+  geometry:       GeometrySpec
+  parties:        Party[]             // parties active in this scenario; min 2
+
+  districts:      District[]          // defines the target district set
+  precincts:      Precinct[]          // editable + context precincts
+
+  group_schema?:  GroupSchema         // optional dimensional schema (see Demographic Groups)
+
+  events:         DemographicEvent[]  // fires at evaluation time
+  rules:          ScenarioRules
+  success_criteria: SuccessCriterion[]
+
+  narrative:      Narrative
+  state_context?: StateContext        // pre-computed results for other regions
+}
+```
+
+---
+
+### `RegionSpec`
+
+```typescript
+interface RegionSpec {
+  state_id:   string    // e.g., "new_texansifornia"
+  region_id:  string    // e.g., "valley_county"
+  name:       string    // display name
+}
+```
+
+---
+
+### `GeometrySpec`
+
+Specifies the coordinate system used by precinct positions. Adjacency is derived
+from coordinates for regular tilings; explicit neighbor lists are required for
+irregular geometries.
+
+```typescript
+type GeometrySpec =
+  | { type: "hex_axial" }
+    // Precincts use axial hex coordinates { q, r }.
+    // All 6 neighbors are derived: (q±1,r), (q,r±1), (q±1,r∓1).
+    // precinct.neighbors must be absent or empty.
+
+  | { type: "custom" }
+    // Precincts use { x, y } cartesian coordinates (for rendering position only).
+    // precinct.neighbors MUST be provided — adjacency is not derivable.
+```
+
+For v1, all scenarios use `"hex_axial"`. `"custom"` is defined now to avoid
+a format-breaking change when organic map shapes are introduced.
+
+---
+
+### `Party`
+
+```typescript
+interface Party {
+  id:           PartyId   // referenced in vote_shares, events, criteria
+  name:         string    // e.g., "Red Party"
+  abbreviation: string    // e.g., "R"; used in UI labels
+}
+```
+
+Parties are entirely fictional. No mapping to real-world parties. All scenarios
+use fictional names. Minimum two parties; N-party supported.
+
+---
+
+### `District`
+
+```typescript
+interface District {
+  id:           DistrictId
+  name?:        string    // display name, e.g., "District 3"
+  description?: string    // optional flavor text
+}
+```
+
+The number of districts in `districts[]` determines how many the player must fill.
+Target population per district is derived: `total_region_population / districts.length`.
+Population balance tolerance is declared in `ScenarioRules.population_tolerance`.
+
+---
+
+### `Precinct`
+
+Note: the map geography ADR declared `neighboring_context_precincts[]` as a separate array. This spec
+merges editable and context precincts into a single `precincts[]` array, distinguished by the `editable`
+boolean. This is a deliberate simplification — a flat list is easier to validate and index than two parallel
+arrays. The semantic distinction (editable vs. context) is preserved via the flag.
+
+```typescript
+interface Precinct {
+  id:                  PrecinctId
+  editable:            boolean     // false = neighboring context precinct (read-only)
+  county_id?:          string      // display only; used to draw county border overlay
+
+  // Geometry
+  position:            HexAxialPosition | CartesianPosition
+  neighbors?:          PrecinctId[]  // required for geometry.type == "custom"; omit for hex_axial
+
+  // Population
+  total_population:    number      // non-negative integer; sum of all group populations
+
+  // Demographics
+  demographic_groups:  DemographicGroup[]  // population_shares must sum to 1.0
+
+  // Starting state
+  initial_district_id?: DistrictId | null
+    // If provided: precinct begins assigned to this district.
+    // If absent/null: precinct is unassigned at scenario start.
+    // When all precincts are absent/null, the game assigns all to districts[0]
+    // as the starting state — player begins with one giant district and carves it up.
+    // See: Unassigned Precinct Handling.
+    //
+    // Note: the map geography ADR specifies `pc.assignments: Map<ElectionType, DistrictId>` as the
+    // runtime data structure. The scenario format uses a bare `initial_district_id` because each
+    // scenario declares exactly one `election_type`; there is no ambiguity about which assignment
+    // layer this initialises. The runtime engine maps `initial_district_id` into the appropriate
+    // slot of the `assignments` Map when loading the scenario.
+
+  // Optional display metadata
+  name?: string         // label for tooltips, e.g., "Riverside Precinct"
+  tags?: string[]       // for precinct filtering in events, e.g., ["urban", "valley"]
+}
+
+interface HexAxialPosition { q: number; r: number }
+interface CartesianPosition { x: number; y: number }
+```
+
+**Context precincts** (`editable: false`) participate in population balance and
+simulation but are not editable. They are first-class in the precinct list; the
+renderer distinguishes them visually.
+
+---
+
+### `DemographicGroup`
+
+```typescript
+interface DemographicGroup {
+  id:               GroupId         // opaque string; unique within precinct
+  population_share: number          // fraction of precinct total_population [0, 1]
+  vote_shares:      Record<PartyId, number>  // must sum to 1.0; all parties present
+  turnout_rate:     number          // fraction of eligible pop that votes [0, 1]
+
+  // Required if scenario declares a group_schema:
+  dimensions?:      Record<DimensionName, DimensionValue>
+    // Must supply a value for every dimension declared in group_schema.dimensions
+}
+```
+
+`vote_shares` must include an entry for every `PartyId` declared in `scenario.parties`.
+
+**Voter eligibility** is NOT a field on groups. It is derived from `group_schema.eligibility_rules`
+(see below). Groups with no applicable eligibility rule are 100% voter-eligible.
+
+**Voter weight** used in simulation:
+```
+voter_weight(group) = total_population × population_share
+                    × voter_eligible(group)          // 1.0 or 0.0
+                    × turnout_rate
+```
+
+---
+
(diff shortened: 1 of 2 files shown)


thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+  neighbors?:          PrecinctId[]  // required for geometry.type == "custom"; omit for hex_axial
+
+  // Population
+  total_population:    number      // non-negative integer; sum of all group populations
+
+  // Demographics
+  demographic_groups:  DemographicGroup[]  // population_shares must sum to 1.0
+
+  // Starting state
+  initial_district_id?: DistrictId | null
```
cgruber, line 161: The map geography ADR explicitly requires `assignments: Map<ElectionType, DistrictId>` on precincts — "Not a single district ID." The spec uses a simpler `initial_district_id?: DistrictId | null`.

The spec's choice is defensible (one ET per scenario, so a bare ID avoids map boilerplate), but it deviates from the ADR's stated shape without acknowledging it. Either:
- Add a short rationale here noting that "one ET per scenario file means the assignment map collapses to a single ID; this is intentional," or
- Update the map geography ADR's Consequences section to align.

Without that acknowledgment, implementers reading both documents will see a contradiction.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+---
+
+### `ScenarioRules`
+
+```typescript
+interface ScenarioRules {
+  population_tolerance:   number    // ±fraction from ideal; e.g., 0.01 = ±1%
+                                    // congressional: typically 0.001; state leg: 0.05–0.10
+  contiguity_required:    boolean   // default true; hard error if violated
+  non_contiguous_allowed: boolean   // default false; if true, disconnected segments valid
```
cgruber: These two booleans express the same constraint with opposite polarity and create a potential contradiction state. When `contiguity_required: true` and `non_contiguous_allowed: true` are both set, the spec says both things at once with no defined winner.

Suggest collapsing to a single field. Either:
- Remove `non_contiguous_allowed` and read `contiguity_required: false` as "non-contiguous maps are allowed," or
- Remove `contiguity_required` and use only `non_contiguous_allowed`.

The island archipelago example only needs one bit; having two is redundant and creates an invalid state the invariants do not currently catch.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+  // Optional display metadata
+  name?: string         // label for tooltips, e.g., "Riverside Precinct"
+  tags?: string[]       // for precinct filtering in events, e.g., ["urban", "valley"]
+}
+
+interface HexAxialPosition { q: number; r: number }
+interface CartesianPosition { x: number; y: number }
+```
+
+**Context precincts** (`editable: false`) participate in population balance and
```
cgruber, line 183: Missing validation invariant: context precincts (`editable: false`) must have a non-null `initial_district_id`.

Context precincts represent already-drawn territory outside the player's editable region. They participate in population balance and simulation — both of which require knowing their district assignment. A context precinct with no `initial_district_id` would be silently auto-assigned to `districts[0]`, potentially corrupting the starting state the scenario author intended.

Suggested invariant to add:
> 11. If `editable: false` on a precinct, `initial_district_id` must be present and non-null.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+   one precinct's `demographic_groups`.
+4. For every precinct: `sum(demographic_groups[*].population_share) == 1.0` (±ε).
+5. For every group: `sum(vote_shares[*]) == 1.0` (±ε); all `scenario.parties` present.
+6. If `group_schema` declared: every precinct contains one group per combination of
+   dimension values; every group declares a value for every dimension.
+7. For `geometry.type == "hex_axial"`: no precinct has a `neighbors` field.
+8. For `geometry.type == "custom"`: every precinct has a non-empty `neighbors` field;
+   adjacency is symmetric (if A lists B as neighbor, B lists A).
+9. `scenario.districts` has at least 2 entries.
+10. Every `EventId`, `CriterionId`, `PrecinctId`, `DistrictId`, `GroupId` is unique
```
cgruber: `PartyId` is missing from the uniqueness invariant. Party IDs are referenced in `vote_shares`, `events`, `success_criteria`, and `state_context`. If two parties share an ID, those references are ambiguous.

The fix is a one-word addition: "Every `EventId`, `CriterionId`, `PrecinctId`, `DistrictId`, `GroupId`, **`PartyId`** is unique within the scenario."

cgruber: Missing invariant: for `geometry.type == "custom"`, neighbor IDs in `precinct.neighbors[]` must reference PrecinctIds that exist in `scenario.precincts`.

The symmetry check (invariant 8) verifies that if A lists B, B lists A — but neither invariant catches a reference to a completely nonexistent precinct ID. Add:

> For `geometry.type == "custom"`: every PrecinctId in `neighbors[]` exists in `scenario.precincts`.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+  id:             ScenarioId
+  title:          string
+  election_type:  ElectionType        // "congressional" | "state_senate" | "state_house"
+
+  region:         RegionSpec
+  geometry:       GeometrySpec
+  parties:        Party[]             // parties active in this scenario; min 2
+
+  districts:      District[]          // defines the target district set
+  precincts:      Precinct[]          // editable + context precincts
```
cgruber, line 56: The map geography ADR says: "Must include `neighboring_context_precincts[]` alongside the editable precinct set" — a separate array. This spec merges everything into a single `precincts[]` with an `editable` flag.

The merged approach is cleaner (one list to iterate, same type, no join needed), but the deviation from the ADR's stated consequence should be noted here or in the ADR. A one-line note — "context precincts are co-located in `precincts[]` distinguished by `editable: false`, collapsing the `neighboring_context_precincts[]` shape from the geography ADR" — would keep the doc history coherent.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+   adjacency is symmetric (if A lists B as neighbor, B lists A).
+9. `scenario.districts` has at least 2 entries.
+10. Every `EventId`, `CriterionId`, `PrecinctId`, `DistrictId`, `GroupId` is unique
+    within the scenario.
+
+---
+
+## Unassigned Precinct Handling
+
+When a scenario provides no `initial_district_id` on any precinct (or all are null),
```
cgruber, line 428: The election simulation ADR says the API shape is `simulate(allPrecincts, assignments, electionType)` — accepting `assignments` as a separate parameter (implying a Map<PrecinctId, DistrictId> or similar). The scenario format doesn't describe how `initial_district_id` maps to runtime assignment state.

This is fine — the scenario format is data, not the runtime API — but it would help the spec include a brief note explaining: "At runtime, the engine derives the `assignments` map from precinct `initial_district_id` values modified by player edits. The scenario file records starting state only."

This closes the gap between the data format spec and the simulation ADR's API description without requiring any structural change.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,576 @@
...
+---
+
+## Open Questions
+
+1. **CVAP vs VAP in `majority_minority` criterion**: The criterion uses CVAP when
+   a citizenship dimension is present. Should the criterion explicitly declare
+   which measure to use (`measure: "cvap" | "vap"`) rather than inferring from
+   schema? More explicit; better for authoring validation.
+
+2. **Event ordering and interaction**: Multiple events applied in sequence may
```
cgruber: Open questions 2 and 3 are answered by the spec itself and read more like rationale for a choice already made than genuine open questions.

- Q2 (event ordering): The spec says "applied in order" — this is a decision, not a question. The open question note reads as "we chose sequential; here's why; revisit if needed," which is better placed as a rationale comment in the §EVENTS section than as an unresolved question.
- Q3 (state_context in v1): The spec includes `state_context?: StateContext` as optional and says "v1: parsed but may be ignored by renderer." The decision is made.

Both should be closed as resolved or rewritten to name the specific remaining uncertainty (e.g., Q2 could ask: "Should event application order be user-visible / documented in narrative? Or is it purely an engine concern?").

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. Added a comment on `initial_district_id` explaining the relationship:

> Note: the map geography ADR specifies `pc.assignments: Map<ElectionType, DistrictId>` as the runtime data structure. The scenario format uses a bare `initial_district_id` because each scenario declares exactly one `election_type`; there is no ambiguity about which assignment layer this initialises. The runtime engine maps `initial_district_id` into the appropriate slot of the `assignments` Map when loading the scenario.

The compressed file §PRECINCT section is updated with an abbreviated version of this note.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. Replaced the redundant boolean pair with a single tri-state enum field:

```typescript
interface ScenarioRules {
  population_tolerance:    number
  contiguity:              "required" | "preferred" | "allowed"
    // "required": non-contiguous districts are a hard error (default)
    // "preferred": non-contiguous triggers a warning but is not an error
    // "allowed": non-contiguous is fully permitted (island scenarios, etc.)
  compactness_threshold?:  number
}
```

The example JSON skeleton in the spec is updated to `"contiguity": "required"`. The compressed §RULES section is updated to match.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. Added as new invariant 4 (old 4–11 shifted accordingly):

> 4. Every context precinct (`editable: false`) must have `initial_district_id` set to a non-null `DistrictId`. Auto-fill to `districts[0]` applies only to editable precincts.

The compressed §VALIDATION section is updated to match (now 11 invariants total after all additions).

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. The uniqueness invariant (now #12 after renumbering) has been updated to include `PartyId`:

> 12. Every `EventId`, `CriterionId`, `PrecinctId`, `DistrictId`, `GroupId`, `PartyId` is unique within the scenario.

The compressed §VALIDATION section is updated to match.

cgruber: Fixed. Added as new invariant 10 (between the existing adjacency symmetry invariant and the district count invariant):

> 10. For `geometry.type == "custom"`: all `PrecinctId` values in `neighbors[]` must exist in `scenario.precincts`.

The compressed §VALIDATION section is updated to match.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. Added a note in the `Precinct` section before the TypeScript interface:

> Note: the map geography ADR declared `neighboring_context_precincts[]` as a separate array. This spec merges editable and context precincts into a single `precincts[]` array, distinguished by the `editable` boolean. This is a deliberate simplification — a flat list is easier to validate and index than two parallel arrays. The semantic distinction (editable vs. context) is preserved via the flag.

The compressed §PRECINCT section is updated with an abbreviated version of this note.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. Added a bridge note at the end of the Unassigned Precinct Handling section:

> The scenario's initial assignments are the starting point for the player's session state. The simulation API `simulate(allPrecincts, assignments, electionType)` receives the *current* assignment map (player's working state), not the initial map. The runtime engine maintains assignment state separately from the scenario file; `initial_district_id` is read once at scenario load.

The compressed §UNASSIGNED_PCS section is updated to reflect this.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
cgruber: Fixed. Closed open questions 2 and 3 as resolved:

**Question 2** (event ordering): Struck through and resolved — events are applied in declaration order (sequential); overlapping effects are order-dependent by design and will be documented in authoring guidance.

**Question 3** (`state_context` in v1): Struck through and resolved — `state_context` is included for forward compatibility; v1 renderer may ignore it; no breaking change needed when v2 state-level view is implemented.

The compressed §OPEN section marks both as `[RESOLVED]` with brief notes.

cgruber:
Re-review after fixes. All 8 original issues are resolved:

1. **initial_district_id / runtime Map note** — present on the Precinct field in both files. ✓
2. **contiguity** — single `"required"|"preferred"|"allowed"` enum with per-value comments. ✓
3. **Context precinct invariant** — validation invariant #4 in both files enforces `editable:false` → non-null `initial_district_id`. ✓
4. **PartyId in uniqueness invariant** — invariant #12 (full) / #11 (compressed) includes `PartyId`. ✓
5. **custom neighbors[] existence check** — invariant #10 (full) / #9 (compressed). ✓
6. **neighboring_context_precincts[] ADR deviation note** — note present before Precinct interface in both files. ✓
7. **simulate() API bridge note** — final paragraph of Unassigned Precinct Handling section (full) and §UNASSIGNED_PCS (compressed). ✓
8. **Open questions 2 and 3** — struck through in full spec, `[RESOLVED]` tagged in compressed. ✓

One minor inconsistency found in the fresh pass — see inline comment. Does not block merge.

LGTM.

thoughts/shared/decisions/2026-04-24-scenario-data-format.md
```diff
@@ -0,0 +1,597 @@
...
+  | { type: "majority_minority";  group_filter: GroupFilter; min_cvap_share: number;
+                                  min_districts: number }
+                                  // at least min_districts must have ≥ min_cvap_share CVAP
+  | { type: "efficiency_gap";     operator: CompareOp; threshold: number }
+  | { type: "mean_median";        party: PartyId; operator: CompareOp; threshold: number }
+  | { type: "compactness";        operator: CompareOp; threshold: number }   // Fraction Kept
+  | { type: "safe_seats";         party: PartyId; margin: number; min_count: number }
+                                  // min_count districts where party wins by ≥ margin
+  | { type: "competitive_seats";  margin: number; min_count: number }
+                                  // min_count districts where margin ≤ threshold
```
cgruber: Minor label inconsistency: the comment says "margin ≤ threshold" but the field in the type definition is `margin`, not `threshold`. The comment should read "min_count districts where the winning margin ≤ margin" (or rephrase as "where the winning margin is within `margin`"). Does not block merge — the intent is clear and the type signature is unambiguous.

cgruber: Fixed: comment now reads "min_count districts where winning margin ≤ margin" — field name used consistently.

Issue #31, opened 2026-04-25
GAME-001: Define scenario TypeScript types from spec


cgruber:
Implements spec-grade TypeScript types for the scenario data format in `game/web/src/model/scenario.ts`.

See thoughts/shared/tickets/GAME-001-scenario-typescript-types.md

cgruber:
All AC met. Merged in PR #33.

PR #34, opened 2026-04-25
sprint1(GAME-002): scenario JSON loader + validator

cgruber:
## Summary

Implements `loadScenario(json: unknown): Scenario` — the boundary between raw JSON and the typed game engine. Everything downstream can trust the result is valid.

- All 13 spec validation invariants enforced with descriptive error messages (identifies which invariant failed and which precinct/group/party ID was involved)
- Auto-fill: editable precincts with absent/null `initial_district_id` resolved to `default_district_id` if set, else `districts[0].id`; returned `Scenario` always has explicit `initial_district_id` on editable precincts
- Context precincts (`editable: false`) must have non-null `initial_district_id` — validated at load time
- 33 unit tests: happy path + one test per invariant + auto-fill variants; hand-rolled TAP runner (no Vitest/Jest in project)
- `game/web/src/model/BUILD.bazel` created as a separate Bazel package: `model_lib`, `loader_lib`, `types_lib`, `generator_lib`, `loader_test`, `loader_typecheck_test`

`bazel test //web/...` — 4/4 passed.

**Note:** `game/web/BUILD.bazel` (owned by CI-002, see #34) lists `types_lib` + `generator_lib` as explicit deps. Both PRs must merge for the full build to be clean independently; merging either one first is safe.

see #31

## Test plan

- [x] `bazel build //web/...` passes
- [x] `bazel test //web/...` passes (all 33 loader tests + typecheck)
- [ ] Verify `loadScenario` rejects a deliberately malformed JSON for each invariant

### DIFF
--- a/game/web/src/model/BUILD.bazel
+++ b/game/web/src/model/BUILD.bazel
@@ -1,0 +1,112 @@
+"""
+Bazel targets for the scenario model:
+
+  model_lib  — scenario.ts (GAME-001 types; DistrictId, PartyId, Scenario, etc.)
+  loader_lib — loader.ts (GAME-002 loadScenario function)
+  types_lib  — types.ts (spike/legacy types for the hex-grid spike renderer)
+  generator_lib — generator.ts (spike/legacy procedural precinct generator)
+
+Creating this BUILD.bazel makes src/model/ a separate Bazel package, so the
+parent's glob(["src/**/*.ts"]) no longer covers this directory.  game/web/BUILD.bazel
+lists all four model targets (model_lib, loader_lib, types_lib, generator_lib) as
+explicit deps so the parent typecheck and bundle remain intact.
+"""
+
+load("@aspect_rules_ts//ts:defs.bzl", "ts_config", "ts_project")
+load("@aspect_rules_js//js:defs.bzl", "js_test")
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
+
+# Standalone ts_config for this package (game/web/src/model/tsconfig.json has no
+# extends field — it is a self-contained config, not derived from //web:tsconfig).
+# A local tsconfig.json avoids requiring visibility changes to //web:tsconfig or
+# //web:tsconfig.json, which are owned by CI-002's BUILD.bazel.
+ts_config(
+    name = "tsconfig",
+    src = "tsconfig.json",
+    visibility = ["//web/src/model:__pkg__"],
+)
+
+# ── types_lib: types.ts (spike/legacy types for the hex-grid renderer) ────────
+
+ts_project(
+    name = "types_lib",
+    srcs = ["types.ts"],
+    tsconfig = ":tsconfig",
+    declaration = True,
+    visibility = ["//web:__pkg__", "//web/src/model:__pkg__"],
+)
+
+# ── generator_lib: generator.ts (spike/legacy procedural precinct generator) ──
+
+ts_project(
+    name = "generator_lib",
+    srcs = ["generator.ts"],
+    tsconfig = ":tsconfig",
+    declaration = True,
+    deps = [":types_lib"],
+    visibility = ["//web:__pkg__", "//web/src/model:__pkg__"],
+)
+
+# ── model_lib: scenario.ts types (GAME-001) ───────────────────────────────────
+
+ts_project(
+    name = "model_lib",
+    srcs = ["scenario.ts"],
+    tsconfig = ":tsconfig",
+    declaration = True,
+    visibility = ["//web:__pkg__", "//web/src/model:__pkg__"],
+)
+
+# ── loader_lib: loadScenario function ────────────────────────────────────────
+
+ts_project(
+    name = "loader_lib",
+    srcs = ["loader.ts"],
+    tsconfig = ":tsconfig",
+    declaration = True,
+    deps = [":model_lib"],
+    visibility = ["//web:__pkg__", "//web/src/model:__pkg__"],
+)
+
+# ── loader_typecheck_test: type-check passes ──────────────────────────────────
+
+build_test(
+    name = "loader_typecheck_test",
+    targets = [":loader_lib"],
+    visibility = ["//visibility:public"],
+)
+
+# ── loader_test_lib: compile the test file ────────────────────────────────────
+
+ts_project(
+    name = "loader_test_lib",
+    srcs = ["loader_test.ts"],
+    tsconfig = ":tsconfig",
+    declaration = True,
+    deps = [":loader_lib", ":model_lib"],
+    testonly = True,
+)
+
+# ── loader_test: run tests with Node built-in test runner ─────────────────────
+#
+# aspect_rules_js js_test runs the entry JS file with Node.js.
+# The compiled output of loader_test_lib is loader_test.js (ES2022 modules).
+# Node 18+ supports ESM natively via --input-type or .mjs; however the compiled
+# output from ts_project is .js with ES module syntax.
+#
+# We pass --experimental-vm-modules (Node ≥ 18) and use the js_test entry_point
+# to point at the compiled .js file.
+
+js_test(
+    name = "loader_test",
+    entry_point = "loader_test.js",
+    data = [
+        ":loader_test_lib",
+        ":loader_lib",
+        ":model_lib",
+    ],
+    node_options = [
+        "--experimental-vm-modules",
+    ],
+    visibility = ["//visibility:public"],
+)
--- a/game/web/src/model/loader.ts
+++ b/game/web/src/model/loader.ts
@@ -1,0 +1,894 @@
+/**
+ * Scenario JSON loader and validator.
+ *
+ * `loadScenario(json: unknown): Scenario` is the single entry point.
+ * It validates all 13 spec invariants, auto-fills missing initial_district_id
+ * on editable precincts, and returns a fully-typed Scenario.
+ *
+ * Every error message names the invariant number and identifies the offending
+ * element (precinct id, group id, party id, etc.) so the caller knows where
+ * the violation is.
+ */
+
+import type {
+  CriterionId,
+  DemographicEvent,
+  DemographicGroup,
+  District,
+  DistrictId,
+  EventId,
+  GroupFilter,
+  GroupId,
+  GroupSchema,
+  Narrative,
+  Party,
+  PartyId,
+  Precinct,
+  PrecinctId,
+  RegionSpec,
+  Scenario,
+  ScenarioId,
+  ScenarioRules,
+  Slide,
+  StateContext,
+  SuccessCriterion,
+  GeometrySpec,
+} from "./scenario.js";
+
+// ─── Epsilon for floating-point sum checks ───────────────────────────────────
+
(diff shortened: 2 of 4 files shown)


game/web/src/model/loader.ts
```diff
@@ -0,0 +1,863 @@
...
+    state_name: requireString(r["state_name"], "state_context.state_name"),
+    total_districts: requireNumber(r["total_districts"], "state_context.total_districts"),
+    other_region_results,
+  };
+}
+
+// ─── Validation helpers ───────────────────────────────────────────────────────
+
+/** Collect all GroupId values referenced in a GroupFilter (only group_ids form; dimension form has no explicit GroupId refs). */
+function groupFilterGroupIds(gf: GroupFilter): GroupId[] {
```
cgruber, line 457: **Invariant 3 / dimension-based GroupFilter is never validated (blocking correctness gap)**

When a `GroupFilter` uses the `{dimension, value}` form, `groupFilterGroupIds` returns `[]`, so the invariant-3 loop skips it entirely. But the spec says this form **requires** `group_schema` (§EVENTS: `{dimension,value}(requires schema)`), and even when a schema is present it never checks that:

1. `group_schema` is actually declared on the scenario
2. The `dimension` name exists in `group_schema.dimensions`
3. The `value` is listed in `group_schema.dimensions[dimension]`

The same gap applies everywhere `parseGroupFilter` is called: `turnout_shift`, `vote_share_shift`, `population_shift`, and `majority_minority` criteria.

Concrete breakage: a scenario with `group_filter: { dimension: "bogus", value: "nonsense" }` and no `group_schema` at all passes all invariants today, then silently produces wrong results (or a null-deref crash) at simulation time.

**Suggested fix** — after parsing each `GroupFilter`, validate dimension-form filters:

```ts
function validateGroupFilter(
  gf: GroupFilter,
  label: string,
  group_schema: GroupSchema | undefined
): void {
  if ("group_ids" in gf) {
    // invariant 3 check: each id must exist in definedGroupIds (caller responsibility)
    return;
  }
  // dimension+value form requires schema
  if (group_schema === undefined) {
    throw new Error(
      `Invariant 3: ${label} uses dimension-based group_filter but no group_schema is declared`
    );
  }
  const allowed = group_schema.dimensions[gf.dimension];
  if (allowed === undefined) {
    throw new Error(
      `Invariant 3: ${label} group_filter dimension "${gf.dimension}" does not exist in group_schema`
    );
  }
  if (!allowed.includes(gf.value)) {
    throw new Error(
      `Invariant 3: ${label} group_filter value "${gf.value}" is not in group_schema.dimensions["${gf.dimension}"]`
    );
  }
}
```

Call it from each place that consumes a `GroupFilter` (events and criteria validation loops). The existing `groupFilterGroupIds` helper and invariant-3 `group_ids` checks stay as-is; this is purely additive.

Also missing: a test covering criterion `majority_minority` with an unknown `group_ids` ref (the code path is there, but untested — the existing Invariant 3 test only covers events).

game/web/src/model/BUILD.bazel
```diff
@@ -0,0 +1,111 @@
...
+parent's glob(["src/**/*.ts"]) no longer covers this directory.  game/web/BUILD.bazel
+lists all four model targets (model_lib, loader_lib, types_lib, generator_lib) as
+explicit deps so the parent typecheck and bundle remain intact.
+"""
+
+load("@aspect_rules_ts//ts:defs.bzl", "ts_config", "ts_project")
+load("@aspect_rules_js//js:defs.bzl", "js_test")
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
+
+# Local ts_config that extends game/web/tsconfig.json via a local tsconfig.json.
```
cgruber: **BUILD comment says tsconfig "extends" parent, but tsconfig.json has no `extends` field**

The comment on the `ts_config` target reads:

> Local ts_config that extends game/web/tsconfig.json via a local tsconfig.json.  
> This avoids requiring visibility changes to //web:tsconfig or //web:tsconfig.json

But `tsconfig.json` as added has no `"extends"` key — it is a fully standalone config, not extending anything. The comment is inaccurate and will mislead anyone who tries to trace the parent config.

Either:
- Add `"extends": "../../tsconfig.json"` (or the correct relative path) to `tsconfig.json` if extension was the intent, **or**
- Update the comment to say it is a standalone config for this package.

(If standalone was intentional, the second bullet is fine — just correct the comment.)

game/web/src/model/loader_test.ts
```diff
@@ -0,0 +1,787 @@
...
+          ],
+          initial_district_id: "d99",
+        },
+      ],
+    })),
+    /Invariant 2.*p1.*d99/
+  );
+});
+
+// ─── Invariant 3: All GroupId refs in events/criteria exist in ≥1 precinct ───
```
cgruber, line 389: **Invariant 3 test gap: no test for criterion `majority_minority` with an unknown `GroupId`**

There is a test that exercises the event-path for Invariant 3 (`event group_filter with unknown GroupId`), but there is no test for a `majority_minority` criterion whose `group_filter` contains an unknown `group_ids` entry. The code handles this path:

```ts
if (c.type === "majority_minority") {
  const gids = groupFilterGroupIds(c.group_filter);
  for (const gid of gids) {
    if (!definedGroupIds.has(gid)) {
      throw new Error(...)
    }
  }
}
```

The AC requires "one test per invariant violation". A single additional test like:

```ts
test("Invariant 3: rejects majority_minority criterion with unknown GroupId", () => {
  assertThrows(
    () => loadScenario(minimalScenario({
      success_criteria: [
        {
          id: "c1",
          required: true,
          description: "Minority rep",
          criterion: {
            type: "majority_minority",
            group_filter: { group_ids: ["g_unknown"] },
            min_eligible_share: 0.5,
            min_districts: 1,
          },
        },
      ],
    })),
    /Invariant 3.*c1.*g_unknown/
  );
});
```

would close this gap.

cgruber, line 389: **Invariant 11 test gap: no cross-precinct duplicate GroupId test**

The spec says GroupIds are unique **within the scenario** (across all precincts), and the implementation correctly checks this globally (all group ids go into the same `allIds` map). But the only test for duplicate GroupIds uses two groups in the *same* precinct.

A cross-precinct case where `g1` appears in both `p1` and `p2` would fail today with "Invariant 11: duplicate id `g1`…" — but it is untested, so a future refactor that changes the scope of the check would not be caught.

Worth adding a short test:

```ts
test("Invariant 11: rejects same group id in two different precincts", () => {
  assertThrows(
    () => loadScenario(minimalScenario({
      precincts: [
        {
          id: "p1", editable: true, position: { q: 0, r: 0 }, total_population: 500,
          demographic_groups: [
            { id: "shared_g", population_share: 1.0, vote_shares: { blue: 0.6, red: 0.4 }, turnout_rate: 0.7 },
          ],
        },
        {
          id: "p2", editable: true, position: { q: 1, r: 0 }, total_population: 500,
          demographic_groups: [
            { id: "shared_g", population_share: 1.0, vote_shares: { blue: 0.6, red: 0.4 }, turnout_rate: 0.7 },
          ],
        },
      ],
    })),
    /Invariant 11.*duplicate id.*shared_g/
  );
});
```

game/web/src/model/loader.ts
cgruber: Fixed. Added a `validateDimensionFilter(gf, schema, label)` helper in `loader.ts` (after `groupFilterGroupIds`, around line 462). It:

- Is a no-op for `group_ids` filters (handled by the existing path).
- For `{dimension, value}` filters, checks:
  - (a) `group_schema` is present — throws `Invariant 3: … uses a dimension GroupFilter but no group_schema is defined`
  - (b) the dimension name exists in `group_schema.dimensions` — throws `Invariant 3: … group_filter references unknown dimension "…"`
  - (c) the value is in `group_schema.dimensions[dimension]` — throws `Invariant 3: … dimension "…" value "…" is not in schema values […]`

`validateDimensionFilter` is called in the Invariant 3 block for every event `group_filter` and for every `majority_minority` criterion `group_filter`, immediately after the `group_ids` existence check loop.

The existing test `"events with dimension group_filter (no group_ids) pass without group existence check"` uses a valid dimension (`age`) and value (`young`) against a schema that declares them — it still passes.

game/web/src/model/BUILD.bazel
cgruber: Fixed. The comment on the `ts_config` rule now reads:

```
# Standalone ts_config for this package (game/web/src/model/tsconfig.json has no
# extends field — it is a self-contained config, not derived from //web:tsconfig).
# A local tsconfig.json avoids requiring visibility changes to //web:tsconfig or
# //web:tsconfig.json, which are owned by CI-002's BUILD.bazel.
```

Removed the false claim about extending the parent config.

game/web/src/model/loader_test.ts
cgruber: Fixed. Added a test after the existing Invariant 3 event test:

```
"Invariant 3: rejects majority_minority criterion with unknown GroupId in group_filter"
```

It uses a `success_criteria` entry with `type: "majority_minority"` and `group_filter: { group_ids: ["nonexistent_group"] }`, asserting the error matches `/Invariant 3.*c1.*nonexistent_group/`. The minimal scenario's precinct only defines group `"g1"`, so `"nonexistent_group"` is absent from all precincts' `demographic_groups`.

cgruber: Fixed. Added a test after the existing Invariant 11 tests:

```
"Invariant 11: rejects duplicate group id across different precincts"
```

It has two distinct precincts (`p1`, `p2`) each with a group named `"shared_group"`. Each precinct has `population_share: 1.0` and `vote_shares` for both parties so no earlier invariant fires. The error must match `/Invariant 11.*duplicate id.*shared_group/`.

This covers the cross-precinct case. The existing test `"Invariant 11: rejects duplicate group ids within a precinct"` covers the same-precinct case.

cgruber:
Re-check after fixes — all four previously raised issues are resolved. LGTM.

**1. BLOCKING (thread 3141751030) — dimension-based GroupFilter validation: RESOLVED**
`validateDimensionFilter` helper is defined at loader.ts:470 and called at both locations in the invariant-3 block: for event group_filters (line 700) and for `majority_minority` criteria group_filters (line 714). Validation covers missing `group_schema`, unknown dimension name, and value not in schema values.

**2. Minor (thread 3141751161) — BUILD.bazel comment re: tsconfig extends: RESOLVED**
Comment now correctly reads "extends field — it is a self-contained config, not derived from //web:tsconfig" (no longer claims it "extends parent").

**3. Minor (thread 3141751298) — Missing test for invariant-3 / majority_minority unknown GroupId: RESOLVED**
Test added at loader_test.ts:407: `"Invariant 3: rejects majority_minority criterion with unknown GroupId in group_filter"`.

**4. Minor (thread 3141751447) — Missing test for invariant-11 cross-precinct duplicate GroupId: RESOLVED**
Test added at loader_test.ts:240: `"Invariant 11: rejects duplicate group id across different precincts"`, asserting `/Invariant 11.*duplicate id.*shared_group/`.

Issue #32, opened 2026-04-25
GAME-004: Extract MapRenderer interface from spike renderer


cgruber:
Extract MapRenderer interface from spike; rename concrete class to SvgMapRenderer.

See thoughts/shared/tickets/GAME-004-map-renderer-interface.md

cgruber:
All AC met. Merged in PR #33.

PR #35, opened 2026-04-25
sprint1(CI-002): Playwright behavioral test harness phase 1

cgruber:
## Summary

Adds the Playwright behavioral test harness (CI-002 Phase 1). Agents and CI can now run `bazel test //web:e2e_test` to get a pass/fail signal from a real browser.

- `game/web/playwright.config.ts` — Chromium only, port 58173, python3 http.server (app is esbuild-bundled statically, no Vite dev server)
- `game/web/e2e/smoke.spec.ts` — verifies `#map-svg` present, at least one `path.hex` element, no console errors on load
- `game/web/e2e_test.sh` — Bazel `sh_test` wrapper: assembles artifacts to tempdir, starts http.server, runs `npx playwright test`, traps EXIT to kill server
- `game/web/BUILD.bazel` — `sh_test` e2e target added; explicit deps on all `//web/src/model` targets including `types_lib` + `generator_lib` (created by GAME-002, see #35)
- `game/package.json` / `game/pnpm-lock.yaml` — `@playwright/test` added as devDependency

**One-time setup before first run** (outside Bazel — browser binaries can't be in the sandbox):
```
cd game && node_modules/.bin/playwright install chromium
```

**Note:** `game/web/src/model/BUILD.bazel` (owned by GAME-002, see #35) must be present for this PR's `//web:app` target to build cleanly. Both PRs are safe to merge independently; the second merge resolves the dep.

see #32 (CI-002 ticket will remain open until Phase 2 is complete after GAME-005)

## Test plan

- [x] `bazel build //web/...` passes
- [x] `bazel test //web/... --test_tag_filters=-e2e` passes
- [ ] `npx playwright install chromium` then `bazel test //web:e2e_test` passes against spike app
- [ ] Smoke test fails if `#map-svg` is removed (regression check)

### DIFF
--- a/game/web/BUILD.bazel
+++ b/game/web/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@aspect_rules_esbuild//esbuild:defs.bzl", "esbuild")
 load("@aspect_rules_ts//ts:defs.bzl", "ts_config", "ts_project")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
 
 ts_config(
     name = "tsconfig",
@@ -8,6 +9,8 @@ ts_config(
 )
 
 # Type-check target: verifies TypeScript types without producing runtime output.
+# Note: src/model/ is a separate Bazel package (owns BUILD.bazel); glob does not
+# cross package boundaries, so model targets are listed explicitly as deps.
 ts_project(
     name = "app_typecheck_lib",
     srcs = glob(["src/**/*.ts"]),
@@ -17,6 +20,10 @@ ts_project(
         "//:node_modules/@types/d3",
         "//:node_modules/zustand",
         "//:node_modules/zundo",
+        "//web/src/model:loader_lib",
+        "//web/src/model:model_lib",
+        "//web/src/model:types_lib",
+        "//web/src/model:generator_lib",
     ],
 )
 
@@ -28,6 +35,8 @@ build_test(
 )
 
 # esbuild bundles TypeScript + npm deps into a single browser-ready JS file.
+# Note: src/model/ is a separate Bazel package; model sources are provided via
+# deps on //web/src/model targets.
 esbuild(
     name = "app",
     srcs = glob(["src/**/*.ts"]),
@@ -36,6 +45,10 @@ esbuild(
         "//:node_modules/d3",
         "//:node_modules/zustand",
         "//:node_modules/zundo",
+        "//web/src/model:loader_lib",
+        "//web/src/model:model_lib",
+        "//web/src/model:types_lib",
+        "//web/src/model:generator_lib",
     ],
     output = "bundle.js",
     format = "esm",
@@ -49,5 +62,34 @@ esbuild(
 filegroup(
     name = "html",
     srcs = ["index.html"],
+    visibility = ["//visibility:public"],
+)
+
+# Playwright behavioral (e2e) smoke test — CI-002 Phase 1.
+#
+# Assembles Bazel-built artifacts into a temp dir, starts python3 http.server
+# on port 58173, and runs npx playwright test against it.
+#
+# local = True: Playwright requires host npx and a Chromium browser binary
+# (installed once via `npx playwright install chromium`). The test cannot run
+# inside the Bazel sandbox.
+#
+# One-time setup outside Bazel:
+#   cd game && npx playwright install chromium
+sh_test(
+    name = "e2e_test",
+    srcs = ["e2e_test.sh"],
+    data = [
+        ":app",
+        ":html",
+        "//rust:wasm_calc_bindgen",
+        "playwright.config.ts",
+        "e2e/smoke.spec.ts",
+    ],
+    local = True,
+    tags = [
+        "no-sandbox",
+        "e2e",
+    ],
     visibility = ["//visibility:public"],
 )
--- a/game/web/e2e_test.sh
+++ b/game/web/e2e_test.sh
@@ -1,0 +1,109 @@
+#!/usr/bin/env bash
+# e2e_test.sh — Bazel sh_test wrapper for Playwright behavioral tests.
+#
+# This script:
+#   1. Assembles Bazel-built artifacts (bundle.js, WASM files, index.html)
+#      into a temp directory using Bazel runfiles.
+#   2. Starts python3 -m http.server on port 58173.
+#   3. Runs npx playwright test against that server.
+#   4. Kills the server on exit (pass or fail).
+#
+# Port 58173 is distinct from the dev serve port (58080) to avoid conflicts.
+# There is no Vite dev server in this project — the app is statically served.
+#
+# Runfiles layout (module name "redistricting_sim" from MODULE.bazel):
+#   $RUNFILES_DIR/redistricting_sim/web/bundle.js
+#   $RUNFILES_DIR/redistricting_sim/web/index.html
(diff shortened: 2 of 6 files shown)


game/web/e2e_test.sh
```diff
@@ -0,0 +1,102 @@
...
+PLAYWRIGHT_CONFIG="${WORKSPACE_DIR}/web/playwright.config.ts"
+if [[ ! -f "${PLAYWRIGHT_CONFIG}" ]]; then
+  echo "ERROR: playwright.config.ts not found at ${PLAYWRIGHT_CONFIG}" >&2
+  exit 1
+fi
+
+# ── Run Playwright ────────────────────────────────────────────────────────────
+# npx resolves playwright from the node_modules in the workspace root (game/).
+cd "${WORKSPACE_DIR}"
+exec npx playwright test --config "web/playwright.config.ts"
```
cgruber: **Bug: `exec` suppresses the cleanup trap — server and temp dir will leak.**

`exec npx ...` replaces the shell process entirely, which means the `trap cleanup EXIT` registered on line ~57 never fires. The python3 http.server will keep running and `SERVE_DIR` will not be removed after the test.

Fix: drop `exec` and propagate the exit code explicitly:

```bash
npx playwright test --config "web/playwright.config.ts"
exit $?
```

This is a blocking issue — flaky port conflicts will appear on repeated `bazel test` runs because the previous server is still listening on 58173.

game/web/e2e_test.sh
```diff
@@ -0,0 +1,102 @@
...
+# Wait for the server to be ready (up to 10 seconds)
+for i in $(seq 1 20); do
+  if python3 -c "import urllib.request; urllib.request.urlopen('http://localhost:${PORT}/')" 2>/dev/null; then
+    break
+  fi
+  sleep 0.5
+done
+
+# ── Resolve workspace root (for playwright.config.ts) ────────────────────────
+# Playwright config is at game/web/playwright.config.ts.
```
cgruber, line 82: **Server readiness loop has no failure path.**

If python3 never successfully binds (e.g., port already taken, binary missing), the loop finishes silently after ~10 s and the script continues. Playwright then times out trying to reach `localhost:58173`, producing a confusing "net::ERR_CONNECTION_REFUSED" error rather than a clear "server did not start."

Add a liveness check immediately after the loop:

```bash
if ! python3 -c "import urllib.request; urllib.request.urlopen('http://localhost:${PORT}/')" 2>/dev/null; then
  echo "ERROR: http.server did not start on port ${PORT} after 10 s" >&2
  exit 1
fi
```

This makes failures fast and diagnostic rather than a 10-second Playwright timeout.

game/web/playwright.config.ts
```diff
@@ -0,0 +1,32 @@
...
+  timeout: 10_000,
+  retries: 0,
+  reporter: [["list"], ["html", { outputFolder: "test-results/html", open: "never" }]],
+  outputDir: "test-results/artifacts",
+
+  use: {
+    baseURL: "http://localhost:58173",
+    // Collect trace on first retry (useful for CI debugging)
+    trace: "on-first-retry",
+  },
```
cgruber, line 24: **`trace: "on-first-retry"` is dead config — `retries: 0` means no retries happen.**

With `retries: 0`, Playwright never executes a first retry, so this trace setting never triggers. The comment above it ("useful for CI debugging") is correct, but the setting defeats itself.

Options:
- Change to `trace: "retain-on-failure"` to capture a trace on any failure without requiring a retry.
- Or set `retries: 1` in CI (can be toggled via env: `retries: process.env.CI ? 1 : 0`).

As written, a failing CI run produces no trace, which is the opposite of what the comment promises.

game/web/BUILD.bazel
```diff
@@ -17,6 +20,10 @@ ts_project(
         "//:node_modules/@types/d3",
         "//:node_modules/zustand",
         "//:node_modules/zundo",
+        "//web/src/model:loader_lib",
```
cgruber, line 23: **Scope question: model dep additions are unrelated to the Playwright harness.**

The four `//web/src/model:*_lib` deps added to `app_typecheck_lib` and `esbuild` fix the app build/typecheck but are not part of the e2e test infrastructure. They appear to have been bundled in to make the test have a working app to test against.

Is this intentional? If these deps were missing before this PR, the app build was already broken in CI. A separate commit (or note in the PR description) would help reviewers understand which change is build-fix vs. test infrastructure.

Not blocking, but worth a call-out for history clarity.

game/web/e2e_test.sh
cgruber: Fixed. Replaced `exec npx playwright test` with `npx playwright test; exit $?` at line 102. The `exec` was replacing the shell process, which prevented the `trap cleanup EXIT` from firing and leaked both the python3 HTTP server process and the `SERVE_DIR` temp directory. The trap now runs correctly on exit whether the tests pass or fail.

game/web/e2e_test.sh
cgruber: Fixed. The readiness loop now sets a `SERVER_READY` flag when the server responds, and after the loop an explicit check prints a clear error and exits 1 if the flag was never set:

```bash
if [[ "${SERVER_READY}" -eq 0 ]]; then
  echo "ERROR: HTTP server on port ${PORT} did not become ready after 10 seconds" >&2
  exit 1
fi
```

This avoids the confusing Playwright network-error timeout that would have occurred previously when python3 failed to bind.

game/web/playwright.config.ts
cgruber: Fixed. Changed `trace: "on-first-retry"` to `trace: "retain-on-failure"`. With `retries: 0` there are no retries, so `on-first-retry` never triggered and traces were never collected. `retain-on-failure` captures a trace on any failing test regardless of retry count, which is what we actually want for CI debugging.

game/web/BUILD.bazel
cgruber: Acknowledged — these deps are intentionally outside the strict CI-002 scope but were a required coordination fix. When GAME-002 introduced `web/src/model/BUILD.bazel`, it created a new Bazel sub-package boundary. Bazel `glob()` calls do not cross sub-package boundaries, so the existing `glob(["src/**/*.ts"])` in the parent `game/web/BUILD.bazel` stopped picking up the model sources. The `//web/src/model:*_lib` deps had to be added explicitly to both `app_typecheck_lib` and the `app` esbuild target to restore a passing build. No code change needed — the deps are correct and necessary.

cgruber:
**Re-review: all previously raised issues resolved — LGTM**

Checked the current diff against each flagged item:

1. **BLOCKING (thread 3141749770) — `exec npx` killed the trap**: Fixed. The final lines are now `npx playwright test --config "web/playwright.config.ts"` followed by `exit $?`. No `exec`, so the `trap cleanup EXIT` fires correctly on both pass and failure.

2. **Should-fix (thread 3141749819) — readiness loop had no failure path**: Fixed. `SERVER_READY=0` is set before the loop, flipped to `1` on success, and a post-loop guard (`if [[ "${SERVER_READY}" -eq 0 ]]; then ... exit 1; fi`) exits 1 if the server never became ready.

3. **Should-fix (thread 3141750087) — `trace: "on-first-retry"` dead with `retries: 0`**: Fixed. `playwright.config.ts` now uses `trace: "retain-on-failure"`, which works regardless of retry count. The config even includes a comment explaining why.

4. **Non-blocking (thread 3141750153) — BUILD.bazel scope question**: Confirmed no code change needed; the explanatory reply was sufficient.

All issues addressed. This is good to merge.

PR #40, opened 2026-04-25
build: add idempotent setup.sh for node_modules + Playwright browser install

cgruber:
## Prerequisites
- [ ] N/A — no parent PR

## Summary
- Adds `game/setup.sh`, an idempotent script that handles the two things Bazel does NOT manage automatically: `node_modules` (via `pnpm install --frozen-lockfile`) and Playwright browser binaries (via `pnpm exec playwright install chromium`).
- Intended for developer onboarding on any new machine/workspace and for CI setup before running `bazel test //...`.
- Bazel manages its own toolchains (TypeScript via rules_ts, Rust 1.85 + wasm32 via rules_rust, wasm-bindgen via rules_rust_wasm_bindgen) — no system `rustup`, `wasm-pack`, or global TypeScript install is needed; those are NOT in scope for this script.
- Prerequisites (node, pnpm, python3, bazel) must already be on PATH; the script checks for them and fails fast with actionable messages rather than attempting installation.

## Validation performed
- [x] Script tested locally — `pnpm install --frozen-lockfile` runs cleanly from a fresh state.
- [x] `chmod +x` applied; file is executable.
- [ ] N/A — no Ansible, kubectl, or SOPS changes.

- Developer machines (new workspace or new machine setup).
- CI (once CI pipeline is configured to call `game/setup.sh` before `bazel test`).

- N/A — no deployment; script is used on-demand by developers and CI.

## Tickets addressed
- No ticket opened for this (hygiene/tooling addition that emerged from sprint setup work).

## Rollback
- Delete `game/setup.sh`. No other files changed.

### DIFF
--- a/game/setup.sh
+++ b/game/setup.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+# setup.sh — Idempotent workspace setup for game/.
+#
+# Handles the two things Bazel does NOT manage automatically:
+#   1. npm dependencies in node_modules/ (used by the Playwright sh_test wrapper)
+#   2. Playwright browser binaries (~/.cache/ms-playwright/)
+#
+# Bazel manages its own toolchains (TypeScript, Rust, wasm-bindgen) via bzlmod.
+# You do not need rustup, wasm-pack, or a system TypeScript install.
+#
+# Prerequisites (not installed by this script — must already be on PATH):
+#   node    >= 18
+#   pnpm    >= 8
+#   python3 (any recent version — used by e2e_test.sh's http.server)
+#   bazel   (bazelisk wired as bazel)
+#
+# Usage:
+#   ./setup.sh          # from any directory
+#   bash game/setup.sh  # from repo root
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+cd "${SCRIPT_DIR}"
+
+# ── Colour helpers ────────────────────────────────────────────────────────────
+green()  { if [[ -t 1 ]]; then printf '\033[0;32m%s\033[0m\n' "$*"; else printf '%s\n' "$*"; fi; }
+yellow() { if [[ -t 1 ]]; then printf '\033[0;33m%s\033[0m\n' "$*"; else printf '%s\n' "$*"; fi; }
+red()    { if [[ -t 1 ]]; then printf '\033[0;31m%s\033[0m\n' "$*"; else printf '%s\n' "$*"; fi; }
+
+# ── Prerequisite checks ───────────────────────────────────────────────────────
+# These are system-level tools the script cannot install.  Fail fast with an
+# actionable message rather than a cryptic error mid-run.
+
+MISSING=()
+
+if ! command -v node >/dev/null 2>&1; then
+  MISSING+=("node (>= 18) — install via nvm, Homebrew, or system package manager")
+fi
+if ! command -v pnpm >/dev/null 2>&1; then
+  MISSING+=("pnpm (>= 8) — install via 'npm install -g pnpm' or 'brew install pnpm'")
+fi
+if ! command -v python3 >/dev/null 2>&1; then
+  MISSING+=("python3 — required by the e2e test HTTP server")
+fi
+if ! command -v bazel >/dev/null 2>&1; then
+  MISSING+=("bazel — install via bazelisk: 'brew install bazelisk' then wire as bazel on PATH")
+fi
+
+if [[ ${#MISSING[@]} -gt 0 ]]; then
+  red "setup.sh: missing prerequisites:"
+  for item in "${MISSING[@]}"; do
+    red "  - ${item}"
+  done
+  exit 1
+fi
+
+# ── Step 1: npm dependencies ──────────────────────────────────────────────────
+# pnpm install is idempotent: it verifies node_modules against pnpm-lock.yaml
+# and exits quickly if everything is up to date.
+
+yellow "==> pnpm install"
+pnpm install --frozen-lockfile
+green "    node_modules up to date"
+
+# ── Step 2: Playwright browser binaries ───────────────────────────────────────
+# Playwright installs browser binaries to ~/.cache/ms-playwright/.
+# This is machine-global (one install covers all workspaces/checkouts).
+# 'playwright install' is idempotent: it skips browsers already at the right version.
+#
+# We use pnpm exec to run the locally-installed playwright (version-locked via
+# pnpm-lock.yaml) rather than a global or npx-fetched version.
+
+yellow "==> playwright install chromium"
+pnpm exec playwright install chromium
+green "    Playwright Chromium ready"
+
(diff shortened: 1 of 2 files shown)


game/setup.sh
```diff
@@ -0,0 +1,82 @@
...
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+cd "${SCRIPT_DIR}"
+
+# ── Colour helpers ────────────────────────────────────────────────────────────
+green() { printf '\033[0;32m%s\033[0m\n' "$*"; }
+yellow() { printf '\033[0;33m%s\033[0m\n' "$*"; }
+red() { printf '\033[0;31m%s\033[0m\n' "$*"; }
```
cgruber: **CI-hostile ANSI color codes — missing TTY guard**

The `green`, `yellow`, and `red` helpers unconditionally emit ANSI escape sequences regardless of whether stdout is a terminal.  In CI (GitHub Actions, etc.) the output ends up littered with raw `\033[0;32m…\033[0m` sequences, which are noisy and can confuse log parsers.

Standard fix — guard on TTY or the `NO_COLOR` convention:

```bash
if [[ -t 1 ]] && [[ "${NO_COLOR:-}" == "" ]]; then
  green()  { printf '\033[0;32m%s\033[0m\n' "$*"; }
  yellow() { printf '\033[0;33m%s\033[0m\n' "$*"; }
  red()    { printf '\033[0;31m%s\033[0m\n' "$*"; }
else
  green()  { printf '%s\n' "$*"; }
  yellow() { printf '%s\n' "$*"; }
  red()    { printf '%s\n' "$*"; }
fi
```

This lets local dev keep colors while CI gets clean output, and respects the widely-adopted `NO_COLOR` env var (https://no-color.org/).

game/setup.sh
```diff
@@ -0,0 +1,82 @@
...
+
+yellow "==> pnpm install"
+pnpm install --frozen-lockfile
+green "    node_modules up to date"
+
+# ── Step 2: Playwright browser binaries ───────────────────────────────────────
+# Playwright installs browser binaries to ~/.cache/ms-playwright/.
+# This is machine-global (one install covers all workspaces/checkouts).
+# 'playwright install' is idempotent: it skips browsers already at the right version.
+#
```
cgruber, line 70: **`pnpm exec playwright` here vs `npx playwright` in `e2e_test.sh` — version mismatch risk**

`setup.sh` (step 2) installs browsers via `pnpm exec playwright install chromium`, which uses the Playwright version locked in `pnpm-lock.yaml`.

But `e2e_test.sh` line 129 runs the tests via:

```bash
npx playwright test --config "web/playwright.config.ts"
```

`npx` will resolve Playwright from `node_modules/.bin` if it exists, but in a Bazel-sandboxed `sh_test` the `node_modules` tree from the workspace is not on PATH — so `npx` falls back to downloading the latest Playwright release from the registry.  That downloaded version may differ from the one whose browser binaries were installed by `setup.sh`, leading to a "Executable doesn't exist" error at runtime or, worse, silent behavioral differences.

The consistent fix is to replace `npx playwright` in `e2e_test.sh` with `pnpm exec playwright` (or an explicit `./node_modules/.bin/playwright`), so both scripts use the same locked version.  That change belongs in a follow-up PR or alongside this one; a TODO comment here would at minimum track the inconsistency.

game/setup.sh
```diff
@@ -0,0 +1,82 @@
...
+fi
+if ! command -v pnpm >/dev/null 2>&1; then
+  MISSING+=("pnpm (>= 8) — install via 'npm install -g pnpm' or 'brew install pnpm'")
+fi
+if ! command -v python3 >/dev/null 2>&1; then
+  MISSING+=("python3 — required by the e2e test HTTP server")
+fi
+
+if [[ ${#MISSING[@]} -gt 0 ]]; then
+  red "setup.sh: missing prerequisites:"
```
cgruber, line 51: **`bazel` listed as a prerequisite but not checked**

The header comment says:

> bazel   (bazelisk wired as bazel)

…but unlike `node`, `pnpm`, and `python3`, `bazel` is not included in the `MISSING` array.  The script's own summary message even ends with `bazel build //...` / `bazel test //...`, so a missing `bazel` will produce a confusing "command not found" after a successful setup rather than an up-front actionable error.

Suggested addition alongside the other checks:

```bash
if ! command -v bazel >/dev/null 2>&1; then
  MISSING+=("bazel — install Bazelisk and wire it as 'bazel' (see https://github.com/bazelbuild/bazelisk)")
fi
```

game/setup.sh
cgruber: Fixed. Color helpers now check `[[ -t 1 ]]` before emitting ANSI escape sequences, falling back to plain `printf '%s\n'` when stdout is not a terminal (CI, pipes, redirects).

game/setup.sh
cgruber: Fixed. Updated `game/web/e2e_test.sh` line 129 to use `pnpm exec playwright` instead of `npx playwright`. This ensures the version-locked playwright from `pnpm-lock.yaml` is used consistently in both setup and test, eliminating the mismatch risk.

game/setup.sh
cgruber: Fixed. Added a `bazel` check to the `MISSING[]` array in `game/setup.sh`, after the `python3` check. The error message includes the bazelisk install instructions consistent with the project convention.

game/setup.sh
cgruber: Fixed. All three color helpers now check `[[ -t 1 ]]` before emitting ANSI sequences, falling back to plain `printf '%s\n'` in non-TTY contexts (CI logs, redirected output).

game/setup.sh
cgruber: Fixed. Both `setup.sh` and `e2e_test.sh` now use `pnpm exec playwright`, ensuring both invoke the same version-locked binary from `pnpm-lock.yaml`.

game/setup.sh
cgruber: Fixed. `bazel` is now checked in the `MISSING[]` loop along with the other prerequisites, with an actionable install hint pointing to bazelisk.

cgruber:
**Round 2 review — LGTM**

All three previously flagged issues are addressed cleanly:

1. **ANSI TTY guard** — `green`/`yellow`/`red` helpers check `[[ -t 1 ]]` and fall back to plain output in non-TTY contexts.
2. **Version-locked Playwright** — both `setup.sh` and `e2e_test.sh` now use `pnpm exec playwright`, ensuring they invoke the same binary pinned in `pnpm-lock.yaml`.
3. **`bazel` prerequisite check** — `bazel` is now in the `MISSING[]` guard with an actionable install hint (bazelisk).

No new issues introduced. The script is idempotent, has correct `set -euo pipefail`, uses `BASH_SOURCE[0]` for robust `SCRIPT_DIR` resolution, and the `--frozen-lockfile` flag prevents lockfile drift. The `serve.sh` reference in the Done section points to an existing file.

PR #42, opened 2026-04-25
ci: add GitHub Actions workflow (Bazel build+test + Playwright e2e)

cgruber:
## Prerequisites
- [ ] N/A — no parent PR

## Summary
- Adds `.github/workflows/ci.yml`: GitHub Actions CI workflow that runs on every push to `main` and every PR targeting `main`.
- Adds `game/.bazelversion` (pins Bazel 9.1.0) so Bazelisk on CI uses the same version as local development.
- Fixes stale `npx` references in `game/web/BUILD.bazel` comments (updated to `pnpm exec playwright`).
- Updates `game/setup.sh` prereq hint from pnpm >= 8 to >= 9 (lockfileVersion 9.0 requires pnpm 9.x).

## What CI does

Single job `build-and-test` on `ubuntu-latest`:
1. Install Bazelisk (reads `.bazelversion` → Bazel 9.1.0)
2. Install pnpm 9 + Node 22
3. Cache: pnpm store (keyed on `pnpm-lock.yaml`), Bazel disk cache (keyed on BUILD/bzl/MODULE files), Bazel repository downloads (keyed on `MODULE.bazel`), Playwright browsers (keyed on `pnpm-lock.yaml`)
4. `./setup.sh` — installs `node_modules/` and Playwright Chromium (idempotent; cache hits make these near-instant)
5. `bazel build //... --config=ci`
6. `bazel test //... --config=ci` — runs TypeScript type-check, unit tests, and the Playwright e2e smoke test

`--config=ci` (from `.bazelrc`) enables sandboxed strategy and the disk cache at `/tmp/bazel-disk-cache-game`.

## Validation performed
- [ ] Workflow YAML reviewed for syntax and step ordering.
- [ ] N/A — cannot run GitHub Actions locally; will verify on first PR push.

- GitHub Actions (CI runners for this repo).
- No local machine changes.

- N/A — workflow activates automatically on merge to main.
- First run will populate the caches (cold start); subsequent runs will be faster.

## Tickets addressed
- No ticket opened (CI hookup is part of the sprint setup work, not separately ticketed).

## Rollback
- Delete `.github/workflows/ci.yml` and `game/.bazelversion`.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      # setup.sh handles the two things Bazel does not manage:
+      #   1. node_modules/ (pnpm install --frozen-lockfile)
+      #   2. Playwright browser binaries (pnpm exec playwright install chromium)
+      # Both steps are idempotent — cache hits make them near-instant.
+      - name: Setup environment
+        run: ./setup.sh
+
+      - name: Bazel build
+        run: bazel build //... --config=ci
```
cgruber, line 85: `bazel test //...` already builds all targets as a prerequisite before running tests — the separate `bazel build //...` step is redundant and doubles the build work.

That said, there is a workflow-UX trade-off: keeping it separate makes build failures show up as a distinct red step in the Actions UI rather than being buried in test output. If that visibility matters to you, keep it. Otherwise, removing it saves time and avoids rebuilding everything twice.

Suggestion: remove the `Bazel build` step, or add a comment explaining why the separate build step is kept intentionally.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
```
cgruber, line 14: No `permissions:` block is declared, so this job inherits the repository's default `GITHUB_TOKEN` permissions. On public repos, the default for same-repo PRs and pushes can include write access to contents, which violates the principle of least privilege.

Since this job only checks out code and runs Bazel/Playwright, it needs no write permissions at all. Add a top-level (or job-level) permissions block:

```yaml
permissions:
  contents: read
```

This is especially important as a habit — if this workflow is ever extended with steps that upload artifacts or post status checks, the permissions block makes the grant explicit rather than inherited.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
```
cgruber, line 8: No `concurrency:` group is defined. Without it, multiple pushes to the same PR branch will queue up and all run to completion, wasting CI minutes on stale commits.

Standard pattern to cancel a previous in-progress run when a new push arrives on the same branch:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

Place this at the top-level of the workflow (alongside `on:` and `jobs:`). `cancel-in-progress: false` is fine for pushes to `main` if you want all main-branch runs to complete, but for PRs, cancellation is almost always the right default.

cgruber:
Replacing with Buildkite pipeline — this project uses Buildkite not GitHub Actions.

PR #43, opened 2026-04-25
ci: wire up Buildkite pipeline (macOS, Bazel build+test, Playwright e2e)

cgruber:
## Prerequisites
- [ ] N/A — no parent PR

## Summary
- Fills in the `.buildkite/pipeline.yml` placeholder (was a stub from before SPIKE-001/002 resolved the stack).
- Adds `game/.bazelversion` (pins Bazel 9.1.0) so Bazelisk uses the same version across agents.
- Fixes stale `npx` references in `game/web/BUILD.bazel` comments → `pnpm exec playwright`.
- Updates `game/setup.sh` prereq hint from pnpm >= 8 to >= 9 (lockfileVersion 9.0 requires pnpm 9.x).
- Deletes stale `.buildkite/scripts/kotlin-integration-test.sh` (Kotlin stack removed).

## What CI does

Two-step pipeline on the macOS agent (`agents: os: macos`, native — no Docker):

**Build** (`key: build`):
1. `cd game && ./setup.sh` — installs `node_modules/` and Playwright Chromium (idempotent; warm agent = near-instant)
2. `bazel build --config=ci //...` — compiles TypeScript, Rust/WASM; fails fast before tests

**Test** (`key: test`, `depends_on: build`):
1. `cd game && ./setup.sh` — re-runs setup (BUILDKITE_CLEAN_CHECKOUT wipes checkout between steps)
2. `bazel test --config=ci //...` — TypeScript type-check, unit tests, Playwright e2e smoke test

`--config=ci` (from `.bazelrc`) enables sandboxed strategy and the Bazel disk cache at `/tmp/bazel-disk-cache-game`. The macOS agent's persistent disk provides warm Bazel and pnpm caches across runs.

Both steps have `retry: automatic: exit_status: "*" limit: 2` (matches polyglot convention).

## Validation performed
- [x] Pipeline YAML reviewed for syntax and dependency graph.
- [ ] N/A — cannot trigger Buildkite locally; will verify on first pipeline run.

- Buildkite macOS agent (reyy or equivalent).
- No Linux agents or Docker containers needed.

- Buildkite picks up `.buildkite/pipeline.yml` automatically on the next build trigger.
- First run will be slower (cold Bazel disk cache, fresh pnpm store); subsequent runs warm.

## Tickets addressed
- No ticket (pipeline was a placeholder stub awaiting stack decision; filling it in is part of Sprint 1 wrap-up).

## Rollback
- Revert `.buildkite/pipeline.yml` to the stub. Delete `game/.bazelversion`.


.buildkite/pipeline.yml
```diff
@@ -1,11 +1,49 @@
...
+    depends_on: build
+    agents:
+      os: macos
+    retry:
+      automatic:
+        - exit_status: "*"
+          limit: 2
+    command: |
+      cd game
+      bazel test --config=ci //...
```
cgruber, line 50: **Bug: `setup.sh` is not called in the test step — Playwright e2e will fail on a cold agent**

`BUILDKITE_CLEAN_CHECKOUT: "true"` wipes the working directory before every step. `node_modules/` and the Playwright Chromium binary that `setup.sh` installs in the build step will not be present when the test step starts — even on the same agent — because the checkout is fresh.

The e2e test (`e2e_test` sh_test, `local = True`) calls `pnpm exec playwright test`, which requires both `node_modules/` and the browser binary. Without them it fails at step startup, not as a flaky test.

`setup.sh` is documented as idempotent and "near-instant on a warm agent", so adding it costs nothing:

```yaml
  - label: ":bazel: Test"
    key: test
    depends_on: build
    agents:
      os: macos
    retry:
      automatic:
        - exit_status: "*"
          limit: 2
    command: |
      cd game
      ./setup.sh        # <-- add this
      bazel test --config=ci //...
```

cgruber:
**Follow-up: `.buildkite/scripts/kotlin-integration-test.sh` references a deleted Kotlin stack**

Not introduced by this PR, but now that the pipeline is live it's worth cleaning up: `.buildkite/scripts/kotlin-integration-test.sh` builds and runs `//kotlin:brackets_client` and `//kotlin:brackets_service`, which no longer exist in this repo. The script is not referenced from the new pipeline, so it won't break CI, but it will mislead future readers.

Suggest removing it (or the whole `.buildkite/scripts/` directory if nothing else lives there) in a follow-up.

.buildkite/pipeline.yml
cgruber: Fixed. Added ./setup.sh to the test step command. BUILDKITE_CLEAN_CHECKOUT wipes the checkout between steps, so setup.sh must run in every step that needs node_modules or Playwright browsers.

cgruber:
Fixed in this PR. Deleted .buildkite/scripts/kotlin-integration-test.sh — the Kotlin stack was removed and the script is no longer referenced.

.buildkite/pipeline.yml
cgruber: Fixed — `./setup.sh` is present in the test step. Thread resolved.

cgruber:
**LGTM** — both previously flagged issues are fixed:

1. `./setup.sh` is now present in the test step, so Playwright e2e will work correctly on a cold agent.
2. `.buildkite/scripts/kotlin-integration-test.sh` has been deleted (stale Kotlin scaffold removed).

Also noted: the additional housekeeping changes (pnpm 8→9 bump in `setup.sh`, new `game/.bazelversion`, comment updates from `npx` to `pnpm exec`) are all clean and consistent with the rest of the PR. No new issues introduced. Ready to merge.

PR #1, opened 2026-04-23
Add game vision and agent team design documents

cgruber:
## Summary

- Adds game vision document (`thoughts/shared/vision/game-vision.md` + `.compressed.md`) covering educational goals, political neutrality framing, core game loop, map mechanics, v1 scenario set, progression model, custom level scope, and v1 scope boundary
- Adds agent team and workflow design research (`thoughts/shared/research/2026-04-23-agent-team-and-workflow-design.md` + `.compressed.md`) covering 10 standing roles, 7 on-demand reviewers, iterative sprint workflow, and NFR ownership
- Updates `AGENTS.md` + `AGENTS.compressed.md` for redistricting-sim purpose; adds `thoughts/shared/vision/` as first entry in layout section
- Adds `AGENT-001` ticket tracking the deferred `$abr` convention fix for compressed docs

## Test plan

- [x] Read `thoughts/shared/vision/game-vision.md` — verify educational goal, neutrality framing, scenario set, and v1 scope boundary are coherent
- [x] Read `thoughts/shared/research/2026-04-23-agent-team-and-workflow-design.md` — verify role roster and workflow cycle are complete
- [x] Confirm `.compressed.md` companions are in sync with their `.md` sources
- [x] Confirm `AGENTS.compressed.md` §LAYOUT lists `vision/` first

Verified by critique agent (3 rounds). Known open issue: `$abr` dollar-prefix convention in compressed docs — tracked in AGENT-001, not blocking.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #44, opened 2026-04-25
build: fix CopyFile darwin-sandbox incompatibility in CI .bazelrc

cgruber:
## Prerequisites
- [ ] N/A — no parent PR

## Summary
- Fixes CI build #1 failure: `CopyFile spawn cannot be executed with any of the available strategies: [darwin-sandbox]`.
- The `ts_config` rule in `aspect_rules_ts` generates a `CopyFile` action that macOS's `darwin-sandbox` cannot execute. Adding `--strategy=CopyFile=local` allows this specific action to run outside the sandbox while keeping everything else sandboxed.
- Root cause: polyglot (the reference for this `.bazelrc` pattern) has no TypeScript, so it never hits this. The CopyFile incompatibility is specific to `aspect_rules_ts`'s `ts_config` rule on macOS.

## Validation performed
- [x] Root cause identified from build log: `game/web/BUILD.bazel:6:10: Copying file web/tsconfig.json failed`.
- [ ] N/A — fix is a one-line `.bazelrc` addition; will verify on next CI run.

- Buildkite macOS agent CI only.

- N/A — `.bazelrc` change takes effect on next build trigger automatically.

## Tickets addressed
- No ticket.

## Rollback
- Remove the `--strategy=CopyFile=local` line from `.bazelrc`.

### DIFF
--- a/game/.bazelrc
+++ b/game/.bazelrc
@@ -1,5 +1,8 @@
 build:ci --spawn_strategy=sandboxed
 build:ci --disk_cache=/tmp/bazel-disk-cache-game
+# CopyFile (used by aspect_rules_ts ts_config) cannot run under darwin-sandbox on macOS.
+# Allow it to fall back to local execution while keeping everything else sandboxed.
+build:ci --strategy=CopyFile=local
 
 # aspect_rules_ts 3.x requires explicit transpiler selection.
 common --@aspect_rules_ts//ts:default_to_tsc_transpiler


game/.bazelrc
```diff
@@ -1,5 +1,8 @@
 build:ci --spawn_strategy=sandboxed
 build:ci --disk_cache=/tmp/bazel-disk-cache-game
+# CopyFile (used by aspect_rules_ts ts_config) cannot run under darwin-sandbox on macOS.
+# Allow it to fall back to local execution while keeping everything else sandboxed.
+build:ci --strategy=CopyFile=local
```
cgruber, line 6: LGTM. The fix is correct and well-scoped.

**Is `--strategy=CopyFile=local` the right fix?** Yes. `CopyFile` is a trivial, deterministic file-copy action generated by `aspect_rules_ts`'s `ts_config` rule. The darwin-sandbox restriction on this action type is a known macOS/Bazel compatibility issue, and `local` execution for `CopyFile` carries essentially zero hermeticity risk.

**Other `aspect_rules_ts` actions that might fail similarly (e.g. `Genrule`, `FileWrite`)?** The CI failure surfaced only `CopyFile`, so the other action types are currently compatible with darwin-sandbox. Worth watching: if a future `aspect_rules_ts` upgrade introduces `FileWrite` or `Genrule` actions, those may need the same treatment. No action needed now.

**Comment accuracy:** The inline comment is accurate — it names the rule, the package, and the intent. No changes needed.

**Risk:** Low. The override is constrained to `build:ci`; local developer builds are unaffected. `CopyFile` is environment-independent so running it locally does not weaken hermeticity meaningfully.

**Deployment/rollback:** Covered in the PR description (N/A for deployment; revert the line to roll back). No concerns.

Approve when the next CI run is green.

Issue #62, opened 2026-04-26
GAME-013: Reset to initial district assignments


cgruber:
## GAME-013: Reset to initial district assignments

Add a "Reset" action that restores all precinct assignments to the scenario's initial state and clears the undo history. Requires inline confirmation gesture to prevent accidental resets.

See ticket: `thoughts/shared/tickets/GAME-013-reset-to-initial.md`

PR #63, opened 2026-04-26
feat(GAME-013): reset to initial district assignments

cgruber:
## Prerequisites
- [N/A] No parent PR dependency

## Summary
- Adds a Reset button in the toolbar that restores all precinct assignments to the scenario's initial state
- Requires inline confirmation gesture ("Reset to start? Yes / No") to prevent accidental resets
- On confirmation: calls `store.resetToInitial()` then `temporalStore.clear()` — map re-renders to initial colors, undo/redo both become disabled

## Validation performed
- [x] `bazel build //web:app_typecheck_lib` — passed, no type errors
- [x] `bazel test //web/src/model:loader_test` — passed

- Browser game only (static frontend)

- N/A — static asset; deployed on next game bundle build

## Tickets addressed
- see #62

## Rollback
- Revert this commit; no data migration involved

### DIFF
--- a/game/web/index.html
+++ b/game/web/index.html
@@ -82,6 +82,60 @@
         border-color: #3a7abd;
       }
 
+      #btn-reset {
+        padding: 4px 10px;
+        background: #3a1020;
+        color: #e0e0e0;
+        border: 1px solid #7a2030;
+        border-radius: 4px;
+        cursor: pointer;
+        font-size: 0.85rem;
+      }
+
+      #btn-reset:hover {
+        background: #5a1828;
+      }
+
+      #reset-confirm {
+        display: none;
+        align-items: center;
+        gap: 6px;
+        font-size: 0.82rem;
+        color: #e0a0a0;
+      }
+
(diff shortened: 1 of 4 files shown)


cgruber:
LGTM. Reviewed GAME-013 reset-to-initial feature:

- **Clone correctness**: `initialAssignments` is captured as `new Map(assignments)` at store creation; `resetToInitial()` does `new Map(initialAssignments)` — the snapshot is never mutated.
- **History clear order**: `resetToInitial()` then `temporalStore.getState().clear()` — correct; the transient zundo entry from `set()` is wiped immediately.
- **Double-click guard**: `btnReset` disables on click; both Yes and No re-enable it. No race condition.
- **Null checks**: All four new DOM refs (`btnReset`, `resetConfirm`, `btnResetConfirm`, `btnResetCancel`) are included in the null-guard block that throws before the IIFE runs.
- **CSS pattern**: `display:none` default with `.visible { display:flex }` override — correct and consistent.

No issues found.

Issue #64, opened 2026-04-26
GAME-010: Map validity panel — population balance, contiguity, unassigned


cgruber:
## GAME-010: Map validity panel — population balance, contiguity, unassigned

Add a live validity panel in the sidebar that shows per-district population balance (±% deviation from ideal), count of unassigned precincts, and contiguity status per district. Updates reactively on every paint action.

See ticket: `thoughts/shared/tickets/GAME-010-map-validity-panel.md`

PR #65, opened 2026-04-26
feat(GAME-010): map validity panel — population balance, contiguity, unassigned

cgruber:
## Prerequisites
- [N/A] No parent PR dependency

## Summary
- Adds a live "Map Validity" panel in the sidebar that updates reactively on every paint action
- New `validity.ts` module: pure `computeValidityStats()` — population balance per district, unassigned count, BFS contiguity check per district
- `renderValidityPanel()` added to `mapRenderer.ts`; reads `scenario.rules.population_tolerance` and `contiguity` mode
- Population balance shows ±% deviation from ideal; colored green/red based on tolerance threshold
- Contiguity status colored red (required) or yellow (preferred) when non-contiguous; hidden when "allowed"

## Validation performed
- [x] `bazel build //web:app_typecheck_lib` — passed, `validity.js` in output
- [x] `bazel test //web/src/model:loader_test` — passed

- Browser game only (static frontend)

- N/A — static asset; deployed on next game bundle build

## Tickets addressed
- see #64

## Rollback
- Revert this commit; no data migration involved

### DIFF
--- a/game/web/index.html
+++ b/game/web/index.html
@@ -201,6 +201,34 @@
         line-height: 1.5;
       }
 
+      .validity-row {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        padding: 4px 6px 4px 10px;
+        border-radius: 4px;
+        font-size: 0.78rem;
+        border-left: 3px solid transparent;
+        background: #0a1f3a;
+        margin-bottom: 3px;
+      }
+
+      .validity-ok { color: #80c080; }
+      .validity-warn { color: #e0c040; }
+      .validity-error { color: #e94560; }
+
+      .validity-section-label {
+        font-size: 0.72rem;
+        color: #606080;
+        text-transform: uppercase;
+        letter-spacing: 0.05em;
+        margin: 6px 0 3px;
+      }
+
+      .validity-badge {
+        font-weight: 600;
+      }
+
       #wasm-status-bar {
         padding: 2px 16px;
         background: #16213e;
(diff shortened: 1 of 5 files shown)


cgruber:
LGTM. Reviewed all focus items:

- **BFS correctness**: `visited.size === inSet.size` correctly detects disconnected components — BFS from seed explores everything reachable; if the count falls short, the district is non-contiguous.
- **`precincts[curr]` guard**: The `undefined` check is safe — the node is added to `visited` before being queued, so a corrupt/phantom ID is skipped without causing a false positive.
- **Assignment iteration**: `for (const [precinctId, distId] of assignments)` is correct `Map` iteration; population sums are accumulated correctly.
- **Tolerance math**: `deviationPct` is in percentage-point units (e.g. 5.0 for 5%); `tolerance * 100` converts the fraction to the same units — comparison is valid. Strict `>` (boundary = ok) is a deliberate design choice, not an off-by-one.
- **`DISTRICT_COLORS[did - 1]`**: Districts are 1-based, array is 0-based — correct, with `?? "#888"` fallback.
- **Contiguity null**: Set to `null` only when `"allowed"`, guarded by `!== null` before rendering — correct.
- **Precinct type usage**: Spike `Precinct` from `types.ts` has `id: number` and `neighbors: (number|null)[]`; BFS guards `nbId !== null` — correct.
- **`ScenarioRules` import**: From `scenario.ts`, exposes `population_tolerance` and `contiguity` — matches usage in both files.

Scope matches GAME-010 exactly.

Non-blocking note: `queue.shift()` in `isContiguous` is O(n) per call. Fine for the spike's small maps. If hex grids grow large, swap to an index pointer (`let head = 0; while (head < queue.length)`) for O(1) dequeue.

Issue #73, opened 2026-04-26
Evaluation phase — Test button, criteria check, pass/fail feedback


cgruber:
Evaluation phase — Test button, criteria check, pass/fail feedback

**Ticket:** [GAME-017](https://github.com/cgruber/redistricting-sim/blob/09361207bbedf74e52fbafa117d50bc14e8d9e17/thoughts/shared/tickets/GAME-017-evaluation-phase.md)

PR #74, opened 2026-04-26
feat(GAME-017): evaluation phase — Submit button, criteria check, pass/fail screen

cgruber:
## Summary

- Adds evaluation phase: Submit button + criteria evaluation + pass/fail result screen
- `evaluate.ts`: pure `evaluateCriteria()` evaluates all `scenario.success_criteria` types; `isMapSubmittable()` gates the Submit button on validity (no unassigned, population balanced, contiguity if required)
- `evaluate_test.ts`: 15 unit tests covering all implemented criterion types and validity gate
- `index.html`: Submit button (disabled until valid); result screen overlay (verdict, per-criterion pass/fail, Keep Drawing / Next Scenario buttons)
- `main.ts`: Submit wired to validity gate; `showResultScreen()` builds and shows result with per-criterion detail
- `e2e/sprint3.spec.ts`: 5 behavioral tests for evaluation phase added
- `simulation/BUILD.bazel`: evaluate targets added
- `web/BUILD.bazel`: evaluate_lib added to typecheck and bundle deps

## Implemented criterion types

| Type | Status |
|---|---|
| `district_count` | ✅ implemented |
| `population_balance` | ✅ implemented |
| `compactness` | ✅ implemented (hex interior-edge fraction) |
| `seat_count` | ✅ implemented |
| `safe_seats` | ✅ implemented |
| `competitive_seats` | ✅ implemented |
| `majority_minority` | stub (returns false + "not implemented") |
| `efficiency_gap` | stub |
| `mean_median` | stub |

Tutorial scenario uses `district_count` (required), `population_balance` (required), and `compactness` (optional) — all three fully implemented.

## Validation performed

- [x] `bazel test //... --test_tag_filters=-e2e` — 9/9 pass
- [x] `bazel build //web:app_typecheck_lib` — clean

- N/A (browser-only)

## Deployment steps

- N/A

## Tickets addressed

- see #73 (GAME-017)

## Rollback

- Revert commit on main; redeploy


game/web/src/main.ts
```diff
@@ -264,6 +292,79 @@ if (wasmEl !== null) {
 		btnReset!.disabled = false;
 	});
 
+	// ── Submit / Evaluation (GAME-017) ────────────────────────────────────────
+	function showResultScreen(pass: boolean) {
```
cgruber: The `pass: boolean` parameter is accepted but never read. `showResultScreen` recomputes `evalResult.overallPass` internally, so the caller's argument is silently ignored. The click handler also redundantly recomputes `validity` just to produce this unused argument:

```ts
showResultScreen(isMapSubmittable(validity, scenario.rules));
```

The `boolean` can be dropped from the signature and the redundant `validity` computation removed from the click handler. This also removes a subtle risk: if the guard in `isMapSubmittable` and the one inside `evaluateCriteria` ever diverge, the displayed verdict could contradict the enabled state of the button.

game/web/src/main.ts
```diff
@@ -264,6 +292,79 @@ if (wasmEl !== null) {
...
+		const validity = computeValidityStats(
+			state.precincts,
+			state.assignments,
+			state.districtCount,
+			scenario.rules,
+		);
+		const evalResult = evaluateCriteria(
+			scenario.success_criteria,
+			validity,
+			state.simulationResult!,
```
cgruber: `state.simulationResult!` uses a non-null assertion on a field typed as `SimulationResult | null`. In practice the store always initialises it eagerly, so this is safe today, but the assertion bypasses TypeScript's null check with no runtime guard. If `simulationResult` is somehow null (e.g. if a future refactor delays initialisation or a scenario loads without running an election), this throws at runtime.

A minimal guard:

```ts
if (!state.simulationResult) return;
const evalResult = evaluateCriteria(..., state.simulationResult, ...);
```

This also makes the early-return consistent with the existing null guard at the top of the function (`if (!resultScreen || ...) return`).

game/web/e2e/sprint3.spec.ts
```diff
@@ -84,3 +109,64 @@ test("intro: objective text is shown from scenario narrative", async ({ page })
...
+  // Initial state: all 30 precincts in district 1, district 2 empty → not submittable
+  await expect(page.locator("#btn-submit")).toBeDisabled();
+});
+
+test("submit: result screen is hidden on initial load", async ({ page }) => {
+  await loadEditor(page);
+  await expect(page.locator("#result-screen")).not.toBeVisible();
+});
+
+test("submit: button enables after painting enough precincts to create a valid 2-district map", async ({ page }) => {
```
cgruber: The test name says "button enables after painting enough precincts to create a valid 2-district map" but the assertion is only `toBeAttached()` with a comment that says "may or may not be enabled." The test body doesn't verify what the name claims.

Either:
- Rename to reflect what is actually tested: `"submit: button remains attached after painting precincts"`, or
- Make it actually assert enabled: paint a strictly balanced valid split (15/15 precincts with balanced population) and assert `not.toBeDisabled()`.

As written, this test cannot catch a regression where the Submit button is removed from the DOM (attach check) but gives no signal about the enable/disable gate logic.

game/web/src/main.ts
```diff
@@ -264,6 +292,79 @@ if (wasmEl !== null) {
...
+			html += `</div>`;
+			html += `<span class="rc-badge">${badge}</span>`;
+			html += `</div>`;
+		}
+		resultCriteriaList.innerHTML = html;
+
+		btnKeepDrawing!.style.display = "";
+		btnNextScenario!.style.display = evalResult.overallPass ? "" : "none";
+
+		resultScreen.classList.remove("hidden");
```
cgruber, line 373: `cr.description` and `cr.detail` are injected raw via `innerHTML`. The data comes from local scenario JSON so the practical XSS surface is low, but if a scenario description ever contains `<`, `>`, or `&` for display purposes (e.g. "R wins ≥ 2 seats") the HTML will mis-render. A `>` in a description string would also open an injection vector if scenario files ever come from an external source.

Prefer building the text nodes with `textContent`:

```ts
const descEl = document.createElement("div");
descEl.className = "rc-desc";
descEl.textContent = cr.description;
```

Only the structural `<div>`/`<span>` wrappers need `innerHTML`; the user-facing text strings can use `textContent` safely.

game/web/src/main.ts
cgruber: Fixed. Dropped the `pass: boolean` parameter from `showResultScreen` and simplified the click handler to just call `showResultScreen()` directly. The redundant `validity` computation in the click handler is also removed.

game/web/src/main.ts
cgruber: Fixed. Added an explicit `if (state.simulationResult === null) return;` guard at the top of `showResultScreen`. The non-null assertion is replaced by a plain reference after the guard narrowed the type.

game/web/e2e/sprint3.spec.ts
cgruber: Fixed. Renamed to "submit: button remains in DOM and is interactive after painting precincts across districts" and cleaned up the comment text to match what the test actually asserts (`toBeAttached`).

game/web/src/main.ts
cgruber: Fixed. Replaced the `innerHTML` string-concatenation loop with explicit `createElement` + `textContent` calls for `cr.description` and `cr.detail`. The list is now populated via `appendChild` after clearing with `innerHTML = ""`.

Issue #76, opened 2026-04-26
Scenario progression — sequential unlock and scenario select screen


cgruber:
Scenario progression — sequential unlock and scenario select screen

**Ticket:** [GAME-018](https://github.com/cgruber/redistricting-sim/blob/09361207bbedf74e52fbafa117d50bc14e8d9e17/thoughts/shared/tickets/GAME-018-progression.md)

cgruber:
All AC met. GAME-018 close-ticket PR (#80) merged.

PR #77, opened 2026-04-26
feat(GAME-018): scenario select screen + progression persistence

cgruber:
## Prerequisites
- [x] Parent PR merged: #74 (GAME-017 evaluation phase)
- [x] Parent PR merged: #75 (GAME-017 close ticket)

## Summary
- Implements GAME-018: scenario select screen + localStorage completion persistence
- New players see the intro screen directly on first load
- Returning players (any completed scenario in localStorage) land on scenario select
- Passing the evaluation marks the scenario complete, saves progress, enables "Next Scenario" → select screen
- Progress uses pure serialization functions (testable without mocking localStorage)
- 18 unit tests for progress.ts serialization/mutation; 5 Playwright e2e tests for progression flows

## Changes
- `game/web/index.html`: `#intro-screen` starts hidden (class="hidden"); JS shows it on startup routing
- `game/web/src/main.ts`:
  - Scenario manifest (hardcoded for v1 with 1 scenario)
  - `startScenarioIntro()` with AbortController for clean re-entry on "Play Again"
  - `renderScenarioCards()` / `showScenarioSelect()` for scenario select screen
  - Progress-aware startup routing: returning player → select, new player → intro
  - `showResultScreen()` saves progress to localStorage on pass
  - "Next Scenario" button routes to scenario select (re-renders cards with updated state)
- `game/web/src/model/progress.ts`: already on main (pure functions + localStorage I/O)
- `game/web/src/model/progress_test.ts`: 18 unit tests (new file)
- `game/web/src/model/BUILD.bazel`: added `progress_test_lib`, `progress_test`
- `game/web/e2e/sprint3.spec.ts`: 5 GAME-018 e2e tests added

## Validation performed
- [x] `bazel build //web:app_typecheck_lib` passes
- [x] `bazel test //web/src/model:progress_test` passes (18 tests)
- [x] `bazel test //web/src/simulation:evaluate_test //web/src/model:loader_test` cached pass

## Tickets addressed
- see #76

### DIFF
--- a/game/web/e2e/sprint3.spec.ts
+++ b/game/web/e2e/sprint3.spec.ts
@@ -1,10 +1,11 @@
 import { test, expect } from "@playwright/test";
 
 /**
- * Sprint 3 behavioral tests — GAME-016 (scenario intro) + GAME-017 (evaluation).
+ * Sprint 3 behavioral tests:
+ *   GAME-016 (scenario intro), GAME-017 (evaluation), GAME-018 (progression).
  *
  * GAME-016: Intro slide flow:
- *   1. Intro screen is visible on initial load (before map editor)
+ *   1. Intro screen is visible on initial load (before map editor) — new player
  *   2. Character info is populated from scenario narrative
  *   3. Slide navigation (Next / Previous) cycles through slides correctly
  *   4. "Start Drawing" appears on the last slide and reveals the editor
@@ -12,10 +13,16 @@ import { test, expect } from "@playwright/test";
  *
  * GAME-017: Evaluation phase:
  *   6. Submit button is disabled on initial (all-D1) state (district 2 empty)
- *   7. Submit button enables after drawing a valid balanced map
- *   8. Submitting a passing map shows the pass result screen
- *   9. Submitting a failing map shows the fail result screen
- *   10. "Keep Drawing" button hides the result screen
+ *   7. Submit button remains in DOM after painting precincts
+ *   8. Clicking submit shows result screen with criteria
+ *   9. "Keep Drawing" button hides the result screen
+ *
+ * GAME-018: Progression:
+ *   10. Scenario select screen is shown for returning players (localStorage has completion data)
+ *   11. Scenario card shows "Completed" status and "Play Again" button for completed scenario
+ *   12. "Play Again" from select screen shows intro then editor
+ *   13. Page reload restores completion state from localStorage
+ *   14. New player (no localStorage) sees intro, not scenario select
  */
 
 /** Navigate, dismiss intro, wait for hex grid. */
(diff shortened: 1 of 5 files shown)


game/web/e2e/sprint3.spec.ts
```diff
@@ -1,21 +1,27 @@
...
- *   10. "Keep Drawing" button hides the result screen
+ *   7. Submit button remains in DOM after painting precincts
+ *   8. Clicking submit shows result screen with criteria
+ *   9. "Keep Drawing" button hides the result screen
+ *
+ * GAME-018: Progression:
+ *   10. Scenario select screen is shown for returning players (localStorage has completion data)
+ *   11. Scenario card shows "Completed" status after completing a scenario
+ *   12. Page reload restores completion state from localStorage
+ *   13. New player (no localStorage) sees intro, not scenario select
```
cgruber: Nit: the GAME-018 doc comment block lists four tests (numbered 10–13), but five are actually implemented. The "Play Again from select screen shows intro then editor" test (line ~213) is missing from the index at the top of the file. Please add it as item 11 (and renumber the existing 11–13 to 12–14, or keep the numbering flat).

game/web/src/main.ts
```diff
@@ -381,9 +389,53 @@ if (wasmEl !== null) {
...
 	});
 
-	// "Next Scenario" is a placeholder until GAME-018 adds the select screen
+	// ── Scenario select screen (GAME-018) ─────────────────────────────────────
+	function renderScenarioCards() {
+		if (!scenarioCardsEl) return;
+		scenarioCardsEl.innerHTML = "";
+		SCENARIO_MANIFEST.forEach((entry, i) => {
+			const completed = isCompleted(progress, entry.id);
+			const unlocked = i === 0 || isCompleted(progress, SCENARIO_MANIFEST[i - 1]!.id);
```
cgruber: Suggestion: with `noUncheckedIndexedAccess` enabled in tsconfig, `SCENARIO_MANIFEST[i - 1]` has type `{ id: string; title: string } | undefined`. The non-null assertion `!` bypasses the exact safety net the project opted into.

The logic is correct at runtime — when `i === 0` the left side of `||` short-circuits so `SCENARIO_MANIFEST[-1]!` is never evaluated, and when `i > 0` the element is always defined — but the assertion suppresses the compiler warning unnecessarily.

A type-safe alternative that degrades correctly:

```ts
const unlocked = i === 0 || isCompleted(progress, SCENARIO_MANIFEST[i - 1]?.id ?? "");
```

An empty-string id will never match any real scenario id, so `isCompleted` returns false → the scenario stays locked. No behaviour change, no assertion.

game/web/src/model/progress_test.ts
```diff
@@ -0,0 +1,155 @@
...
+	if (condition) {
+		console.log(`ok ${total} - ${description}`);
+		passed++;
+	} else {
+		console.log(`not ok ${total} - ${description}`);
+		failed++;
+	}
+}
+
+function assertEqual<T>(description: string, actual: T, expected: T): void {
```
cgruber: Nit: `assertEqual` is defined here but never called anywhere in the file — `assertDeepEqual` is used for all structural checks instead. `noUnusedLocals` isn't set so this compiles fine, but it's dead code. Either add a test that uses it (e.g., testing that `serializeProgress` returns a specific exact string) or remove the function.

cgruber:
LGTM with nits. The GAME-018 implementation is well-structured:

- AbortController refactor correctly prevents duplicate intro listeners on replay — a genuine improvement over the manual `escHandler` cleanup.
- `progress = markCompleted(...)` re-assignment is valid (`let` at correct scope).
- `seedProgress()` via `addInitScript` is the right Playwright pattern; it re-runs on `page.reload()` as the reload test relies on.
- The `i === 0` short-circuit makes the `SCENARIO_MANIFEST[i - 1]!.id` safe at runtime (see inline suggestion for a cleaner type-safe form).
- 18 unit tests in `progress_test.ts` cover all four exported functions including edge cases (malformed JSON, null JSON, non-string array items, idempotency, immutability).
- 5 e2e tests cover the full progression flow (new player, returning player, card status, reload persistence, Play Again path).

Three inline nits posted — none are blocking.

game/web/e2e/sprint3.spec.ts
cgruber: Fixed. Updated the GAME-018 section in the doc comment to list all 5 tests (10–14), including "Play Again from select screen shows intro then editor".

game/web/src/main.ts
cgruber: Fixed. Changed to `SCENARIO_MANIFEST[i - 1]?.id ?? ""`. The empty string is safe: `isCompleted(progress, "")` always returns false since no scenario has an empty-string id, so the unlock chain behaves identically while satisfying `noUncheckedIndexedAccess`.

game/web/src/model/progress_test.ts
cgruber: Fixed. Removed the unused `assertEqual` function; only `assertDeepEqual` and `assert` are needed for the progress test cases.

Issue #99, opened 2026-04-27
Player progress persistence — save/resume game state


cgruber:
Player progress persistence — save/resume game state

**Ticket:** [GAME-007](https://github.com/cgruber/redistricting-sim/blob/c10c17dee329af2b85c42d0605dd5098092fb9dc/thoughts/shared/tickets/GAME-007-player-progress-persistence.md)

PR #100, opened 2026-04-27
feat(GAME-007): in-progress scenario save/resume via localStorage WIP

cgruber:
## Prerequisites
- [x] N/A — standalone feature, no parent PR dependency

## Summary
- Add in-progress (WIP) scenario save/resume so players can quit and return without losing work
- `progress.ts`: new `WipState` type + `saveWip` / `loadWip` / `clearWip` functions; storage key `redistricting-sim-wip`; also rename internal `STORAGE_KEY` → `PROGRESS_KEY` (no behavior change)
- `gameStore.ts`: new `restoreAssignments(assignments, activeDistrict)` action
- `main.ts`: restore WIP on scenario load (with undo-history clear); debounced 800ms WIP save on every assignment change; clear WIP on scenario completion; scenario select cards show "In Progress" badge + "Continue" button when WIP exists; routing shows select screen when WIP exists (not only when completed)
- `sprint3.spec.ts`: 4 new e2e tests — WIP badge visible, select screen shown for WIP-only player, assignment map restored, WIP cleared after pass

## Validation performed
- [x] `bazel build //web/...` — clean (no type errors)
- [x] `bazel test //web:e2e_test` — 52/52 pass

- N/A — static browser bundle only; all WIP state is client-side (localStorage); no server-side component

- N/A — static bundle; no server restart or migration required

## Tickets addressed
- see #99

## Rollback
- Revert the 4-file commit; leftover `redistricting-sim-wip` localStorage keys are harmless

### DIFF
--- a/game/web/e2e/sprint3.spec.ts
+++ b/game/web/e2e/sprint3.spec.ts
@@ -255,6 +255,137 @@ test("progression: new player (no localStorage) sees intro, not scenario select"
   await expect(page.locator("#scenario-select")).not.toBeVisible();
 });
 
+// ─── GAME-007: WIP save/resume ───────────────────────────────────────────────
+
+/** Seed a WIP save in localStorage before the page loads. */
+async function seedWip(
+  page: import("@playwright/test").Page,
+  scenarioId: string,
+  assignments: Record<string, number>,
+  activeDistrict: number,
+): Promise<void> {
+  await page.addInitScript(
+    (args: { scenarioId: string; assignments: Record<string, number>; activeDistrict: number }) => {
+      localStorage.setItem("redistricting-sim-wip", JSON.stringify(args));
+    },
+    { scenarioId, assignments, activeDistrict },
+  );
+}
+
+test("wip: scenario select shows 'In Progress' for scenario with saved WIP", async ({ page }) => {
+  // Seed completion for tutorial-002 so select screen appears, plus WIP for scenario-002
+  await seedProgress(page, ["tutorial-002"]);
+  await seedWip(page, "scenario-002", { "0": 1, "1": 2 }, 2);
+  await page.goto("/");
+
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  // scenario-002 card must show "In Progress" status and "Continue" button
+  const cards = page.locator(".scenario-card");
+  const scenario002Card = cards.nth(1); // second card in manifest
+  await expect(scenario002Card.locator(".sc-status.in-progress")).toBeVisible();
+  await expect(scenario002Card.locator(".sc-status.in-progress")).toContainText("In Progress");
+  await expect(scenario002Card.locator(".sc-play-btn.continue")).toBeVisible();
+  await expect(scenario002Card.locator(".sc-play-btn.continue")).toContainText("Continue");
+});
+
+test("wip: select screen shown when WIP exists even for new player", async ({ page }) => {
+  // No completed scenarios, but a WIP exists — should show select screen
+  await seedWip(page, "tutorial-002", { "0": 1 }, 1);
+  await page.goto("/");
+
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  await expect(page.locator("#intro-screen")).not.toBeVisible();
+});
+
+test("wip: saved assignment map is restored on scenario load", async ({ page }) => {
+  // Pre-paint precinct 0 → district 2, precinct 1 → district 3 in the WIP
+  // Then load tutorial-002 and verify via window.__gameStore that assignments match
+  const wipAssignments: Record<string, number> = {};
+  // tutorial-002 has 196 precincts; build a complete assignment with index 0 → 2 (district 2)
+  for (let i = 0; i < 196; i++) wipAssignments[String(i)] = i === 0 ? 2 : 1;
+
+  await seedWip(page, "tutorial-002", wipAssignments, 2);
+  await page.goto("/?s=tutorial-002");
+
+  // Wait for editor to be ready (intro appears because it's a fresh load without completed state)
+  const skip = page.locator("#btn-intro-skip");
+  await expect(skip).toBeVisible({ timeout: 10_000 });
+  await skip.click();
+  await expect(page.locator("path.hex").first()).toBeVisible({ timeout: 10_000 });
+
+  // Verify that precinct 0 is assigned to district 2 via the exposed store
+  const district = await page.evaluate(() => {
+    const store = (window as unknown as Record<string, unknown>)["__gameStore"] as
+      | { getState: () => { assignments: Map<number, number> } }
+      | undefined;
+    return store?.getState().assignments.get(0);
+  });
+  expect(district).toBe(2);
+});
+
+test("wip: painting precincts triggers a debounced WIP save to localStorage", async ({ page }) => {
+  // Load tutorial-002, skip intro, paint precinct 0 → district 2 via store shortcut,
+  // then wait > 800ms and confirm the WIP key was written with the expected assignment.
+  await page.goto("/?s=tutorial-002");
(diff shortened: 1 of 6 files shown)


game/web/src/main.ts
```diff
@@ -464,7 +512,10 @@ if (
 	});
 
 	// ── Subscribe to state changes ────────────────────────────────────────────
-	store.subscribe(() => updateUI());
+	store.subscribe(() => {
+		updateUI();
+		scheduleWipSave();
```
cgruber, line 523: **Bug (minor / wasteful): `scheduleWipSave()` fires during the WIP restore itself**

`restoreAssignments()` calls `set()`, which synchronously notifies all Zustand subscribers. That means `scheduleWipSave()` is called during the restore block — before `temporalStore.getState().clear()` runs. 800 ms later, `saveWip()` writes back the same state that was just loaded. It's not *incorrect* today (the data is identical), but:

1. It's a spurious write on every page load that has a saved WIP.
2. If any future code path mutates state between the restore `set()` and the 800 ms timer firing, the timer could clobber the pending change.
3. It's simply surprising — the debounce is meant to save player-driven changes, not echo restorations.

Suggested fix — guard `scheduleWipSave()` with a flag:

```ts
let restoringWip = false;

// ... restore block ...
restoringWip = true;
store.getState().restoreAssignments(restoredMap, savedWip.activeDistrict);
temporalStore.getState().clear();
restoringWip = false;

// ... scheduleWipSave ...
function scheduleWipSave() {
  if (restoringWip) return;
  // ... rest unchanged
}
```

Since `restoreAssignments` is synchronous, the flag is set and cleared before any microtask/macrotask can observe it, so this is race-free.

game/web/e2e/sprint3.spec.ts
```diff
@@ -255,6 +255,104 @@ test("progression: new player (no localStorage) sees intro, not scenario select"
...
+    store.getState().paintStroke([70, 71, 72, 73, 74], 1);
+  });
+
+  await expect(page.locator("#btn-submit")).toBeEnabled({ timeout: 3_000 });
+  await page.locator("#btn-submit").click();
+  await expect(page.locator("#result-screen")).toBeVisible();
+  await expect(page.locator("#result-verdict")).toHaveText("Map Passed!");
+
+  // WIP key must be gone from localStorage after a successful pass
+  const wipAfter = await page.evaluate(() => localStorage.getItem("redistricting-sim-wip"));
```
cgruber, line 385: **Missing test coverage: the debounce-save path itself is never exercised**

All 4 new tests verify *restore* (seeding WIP via `addInitScript` and checking it loads) and *clear* (submit-pass removes the key). None verify that *playing the game* — actually painting precincts — writes WIP to `localStorage`.

The debounce-save logic in `main.ts` (the `scheduleWipSave` / `saveWip` path triggered by `store.subscribe`) is the core of the AC:

> "When the player is in an active scenario, their current district assignment map is saved to storage on every meaningful state change (debounced)"

This is entirely untested by the PR's e2e suite. A test like the one below would close the gap:

```ts
test("wip: painting precincts saves WIP to localStorage within debounce window", async ({ page }) => {
  await page.goto("/?s=tutorial-002");
  const skip = page.locator("#btn-intro-skip");
  await expect(skip).toBeVisible({ timeout: 10_000 });
  await skip.click();
  await expect(page.locator("path.hex").first()).toBeVisible({ timeout: 10_000 });

  // Drive a paint stroke through the exposed store
  await page.evaluate(() => {
    const store = (window as unknown as Record<string, unknown>)["__gameStore"] as
      | { getState: () => { paintStroke: (ids: number[], d: number) => void } }
      | undefined;
    store?.getState().paintStroke([0], 2);
  });

  // Wait for the 800 ms debounce + a small margin
  await page.waitForTimeout(1_000);

  const wipRaw = await page.evaluate(() => localStorage.getItem("redistricting-sim-wip"));
  expect(wipRaw).not.toBeNull();
  const wip = JSON.parse(wipRaw!);
  expect(wip.scenarioId).toBe("tutorial-002");
  expect(wip.assignments["0"]).toBe(2);
});
```

The AC for this ticket cannot be considered verified without this test. Blocking.

game/web/src/main.ts
```diff
@@ -128,12 +139,14 @@ if (
...
-			statusEl.textContent = completed ? "Completed" : unlocked ? "Ready" : "Locked";
+			const statusLabel = completed ? "Completed" : inProgress ? "In Progress" : unlocked ? "Ready" : "Locked";
+			const statusClass = completed ? "completed" : inProgress ? "in-progress" : unlocked ? "unlocked" : "locked";
+			statusEl.className = `sc-status ${statusClass}`;
+			statusEl.textContent = statusLabel;
 
 			const playBtn = document.createElement("button");
-			playBtn.className = `sc-play-btn ${completed ? "replay" : unlocked ? "play" : "locked-btn"}`;
-			playBtn.textContent = completed ? "Play Again" : unlocked ? "Play" : "Locked";
+			playBtn.className = `sc-play-btn ${completed ? "replay" : inProgress ? "continue" : unlocked ? "play" : "locked-btn"}`;
```
cgruber, line 148: **CSS gap: `.sc-status.in-progress` and `.sc-play-btn.continue` have no style rules**

`index.html` (the single `<style>` block, lines ~374–392) defines rules for every variant of these two classes *except* the two added by this PR:

```css
/* present */
.sc-status.completed { background: #1a4a30; color: #4caf80; }
.sc-status.unlocked  { background: #0f3460; color: #80a8d8; }
.sc-status.locked    { background: #222230; color: #505070; }

.sc-play-btn.play    { background: #e94560; color: #fff; }
.sc-play-btn.replay  { background: #1a4a30; color: #4caf80; border: 1px solid #2a6040; }
.sc-play-btn.locked-btn { background: #222230; color: #404060; cursor: not-allowed; }

/* MISSING */
/* .sc-status.in-progress { ... } */
/* .sc-play-btn.continue  { ... } */
```

Without these rules, the "In Progress" status badge will render with whatever the base `.sc-status` styles produce (unstyled/default background), and the "Continue" button will look identical to a plain `<button>` — no color differentiation from "Play".

The e2e test `"wip: scenario select shows 'In Progress' for scenario with saved WIP"` only checks for `toBeVisible()` and `toContainText()` — it doesn't assert any visual styling — so it will pass even with the styles missing.

Suggested additions (palette TBD — something warm/amber to distinguish from "Ready" blue and "Completed" green):

```css
.sc-status.in-progress { background: #4a3010; color: #d4a030; }
.sc-play-btn.continue  { background: #d4a030; color: #1a1a2e; }
.sc-play-btn.continue:hover { background: #b88020; }
```

Blocking — the feature is visually broken without these rules.

game/web/src/main.ts
cgruber: Fixed. Added `let isRestoringWip = false` before the restore block, and `scheduleWipSave()` now guards `if (isRestoringWip) return` as its first line. Since `store.subscribe` is wired later anyway, the guard is primarily defensive against future reordering.

game/web/e2e/sprint3.spec.ts
cgruber: Fixed. Added a new test "wip: painting precincts triggers a debounced WIP save to localStorage" that paints precinct 0 → district 2 via `__gameStore`, waits 1100ms, then asserts `localStorage.getItem("redistricting-sim-wip")` is non-null and contains `assignments["0"] === 2`.

game/web/src/main.ts
cgruber: Fixed. Added to `index.html`:
- `.sc-status.in-progress { background: #3a2800; color: #d4a030; }` (amber badge)
- `.sc-play-btn.continue { background: #6a4a00; color: #f0c050; border: 1px solid #a07010; }` + hover rule

Amber palette chosen to visually distinguish from "Ready" (blue) and "Completed" (green).

PR #162, opened 2026-04-30
feat(GAME-048): campaign-driven scenario select

cgruber:
feat(GAME-048): campaign-driven scenario select

Adds ?campaign=<id> URL parameter support to scenario select screen. Filters
scenario list to campaign's scenarioIds in campaign order; falls back to all 9
scenarios when absent/unknown. Back button visible only when campaign is active;
navigates to ./?view=campaigns. Calls saveLastPlayedScenario() on scenario load.

Also fixes 4 critique issues (campaign param leak in URLs, WIP confirm handler,
unknown-campaign test gap, Back button aria-label) and patches release.main.kts
to append ?v=<version> to bundle.js at deploy time (cache-busting kludge until
BUILD-009 lands).

### DIFF
--- a/game/release.main.kts
+++ b/game/release.main.kts
@@ -339,6 +339,24 @@ class Deploy : CliktCommand(
             }
             extractZip(stagedZip, deployRoot)
 
+            // Patch index.html: add version query string to bundle.js src so browsers
+            // treat each deploy as a unique resource and bypass stale cache entries.
+            // (BUILD-009 tracks the proper content-hash solution.)
+            val indexHtml = File(deployRoot, "index.html")
+            if (indexHtml.exists()) {
+                val original = indexHtml.readText()
+                val patched = original.replace(
+                    """s.src = "bundle.js";""",
+                    """s.src = "bundle.js?v=$version";"""
+                )
+                if (patched == original) {
+                    System.err.println("⚠ WARNING: cache-bust patch had no effect — 's.src = \"bundle.js\";' not found in index.html")
+                    System.err.println("  Browsers may serve a stale bundle. Check index.html and update release.main.kts.")
+                } else {
+                    indexHtml.writeText(patched)
+                }
+            }
+
             val deployMeta = DeployMetadata(
                 version = version,
                 commit = meta.commitId,
--- a/game/web/e2e/scenarios.spec.ts
+++ b/game/web/e2e/scenarios.spec.ts
@@ -881,3 +881,45 @@ test("about page: accessible from select screen and shows educational content", 
   await expect(page.locator("#scenario-select")).toBeVisible();
   await expect(page.locator("#about-screen")).not.toBeVisible();
 });
+
+// ─── GAME-048: Campaign-driven scenario select ──────────────────────────────
+
+test("campaign select: ?campaign=tutorial shows only tutorial-001 and tutorial-002", async ({ page }) => {
+  await page.goto("/?campaign=tutorial");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  const cards = page.locator(".scenario-card");
+  await expect(cards).toHaveCount(2);
+  await expect(cards.nth(0)).toContainText("Welcome to Redistricting");
+  await expect(cards.nth(1)).toContainText("Three-District Challenge");
+});
+
+test("campaign select: ?campaign=tutorial Back button is visible", async ({ page }) => {
+  await page.goto("/?campaign=tutorial");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  await expect(page.locator("#btn-back-to-campaign")).toBeVisible();
+});
+
+test("campaign select: Back button absent without ?campaign= param", async ({ page }) => {
+  await page.goto("/");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  await expect(page.locator("#btn-back-to-campaign")).not.toBeVisible();
+});
+
+test("campaign select: no ?campaign= shows all 9 scenarios (fallback regression guard)", async ({ page }) => {
+  await page.goto("/");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  const cards = page.locator(".scenario-card");
+  const count = await cards.count();
+  expect(count).toBe(9);
+});
+
+test("campaign select: unknown ?campaign= falls back to all scenarios with Back button hidden", async ({ page }) => {
+  await page.goto("/?campaign=bogus");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  // Unknown campaign → full manifest fallback (9 cards)
+  const cards = page.locator(".scenario-card");
+  const count = await cards.count();
+  expect(count).toBe(9);
+  // Back button must not be visible — bogus param must not leak
+  await expect(page.locator("#btn-back-to-campaign")).not.toBeVisible();
+});
--- a/game/web/index.html
+++ b/game/web/index.html
@@ -12,6 +12,7 @@
   <body>
     <!-- ── Scenario select screen (GAME-018) ────────────────────────────── -->
     <div id="scenario-select" class="hidden" role="region" aria-label="Scenario Selection">
+      <button id="btn-back-to-campaign" hidden aria-label="Back to campaign select" style="background:none;border:1px solid #2a5a8c;color:#8898b0;padding:4px 12px;border-radius:4px;cursor:pointer;font-size:0.8rem;margin-bottom:8px;">← Back</button>
       <h1>Past the Post</h1>
       <div id="scenario-cards" role="list"></div>
       <div style="display:flex;gap:16px;margin-top:16px;align-items:center;">
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -30,6 +30,7 @@ import {
 	clearWip,
 	type WipState,
 } from "./model/progress.js";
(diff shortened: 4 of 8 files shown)


game/web/src/main.ts
```diff
@@ -171,7 +196,10 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 						// Warn: switching will discard in-progress work on a different scenario
 						showWipWarning(currentWip.scenarioId, entry.id);
 					} else {
-						window.location.assign(`./?s=${entry.id}`);
+						const dest = campaignParam
+							? `./?s=${entry.id}&campaign=${campaignParam}`
```
cgruber, line 200: **Bug: unknown `?campaign=` value leaks into card navigation URLs**

When the user visits `/?campaign=bogus` (an unrecognized campaign ID):

- `campaignParam = "bogus"` (truthy)
- `activeCampaign = undefined` (getCampaign returned nothing)
- `activeList` correctly falls back to the full manifest

But the card click destination is gated on `campaignParam`, not `activeCampaign`:

```ts
const dest = campaignParam
  ? `./?s=${entry.id}&campaign=${campaignParam}`  // fires for bogus too
  : `./?s=${entry.id}`;
```

So every card navigates to `./?s=foo&campaign=bogus`, carrying the bogus param forward. The same bug exists in the WIP-confirm handler (`showWipWarning` → `onConfirm`).

**Fix:** derive an effective param from the resolved campaign:

```ts
const effectiveCampaignParam = activeCampaign ? campaignParam : "";
```

Then use `effectiveCampaignParam` in both `dest` constructions.

game/web/src/main.ts
```diff
@@ -195,7 +223,10 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		const onConfirm = () => {
 			cleanup();
 			clearWip();
-			window.location.assign(`./?s=${targetId}`);
+			const dest = campaignParam
+				? `./?s=${targetId}&campaign=${campaignParam}`
```
cgruber, line 227: **Same campaign-param leak in WIP confirm handler**

The `onConfirm` callback inside `showWipWarning` has the same issue as the card click handler: it gates on `campaignParam` (truthy for any non-empty string) rather than `activeCampaign` (only truthy when the campaign ID is recognized):

```ts
const dest = campaignParam
  ? `./?s=${targetId}&campaign=${campaignParam}`  // fires even for unknown IDs
  : `./?s=${targetId}`;
```

After the fix in the card-click handler (`effectiveCampaignParam`), use that same variable here for consistency.

game/web/e2e/scenarios.spec.ts
```diff
@@ -881,3 +881,34 @@ test("about page: accessible from select screen and shows educational content",
...
+  await expect(page.locator("#btn-back-to-campaign")).not.toBeVisible();
+});
+
+test("campaign select: no ?campaign= shows all 9 scenarios (fallback regression guard)", async ({ page }) => {
+  await page.goto("/");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  const cards = page.locator(".scenario-card");
+  const count = await cards.count();
+  expect(count).toBe(9);
+});
```
cgruber, line 914: **Test coverage gap: unknown `?campaign=` value not tested**

Test #4 covers the no-param fallback (9 cards at `/`). But there's no test for an unrecognized campaign ID (e.g. `/?campaign=bogus`). This would have caught the URL-leak bug in comment #1.

Suggested addition:

```ts
test("campaign select: unknown ?campaign= falls back to all scenarios with clean card URLs", async ({ page }) => {
  await page.goto("/?campaign=bogus");
  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
  const cards = page.locator(".scenario-card");
  await expect(cards).toHaveCount(9);
  // Back button should be hidden — no recognized campaign
  await expect(page.locator("#btn-back-to-campaign")).not.toBeVisible();
});
```

The href/URL assertion for cards is harder in Playwright (cards use click handlers, not `<a>` tags), but at minimum the count and Back-button visibility checks would confirm the fallback path.

game/web/index.html
```diff
@@ -12,6 +12,7 @@
   <body>
     <!-- ── Scenario select screen (GAME-018) ────────────────────────────── -->
     <div id="scenario-select" class="hidden" role="region" aria-label="Scenario Selection">
+      <button id="btn-back-to-campaign" hidden style="background:none;border:1px solid #2a5a8c;color:#8898b0;padding:4px 12px;border-radius:4px;cursor:pointer;font-size:0.8rem;margin-bottom:8px;">← Back</button>
```
cgruber: **Accessibility: Back button `←` arrow is read inconsistently by screen readers**

The raw U+2190 left-arrow character `←` in button text is announced differently across screen readers (VoiceOver says "left arrow", JAWS may skip it entirely). Add an `aria-label` so the accessible name is unambiguous:

```html
<button id="btn-back-to-campaign" hidden aria-label="Back to campaign select" style="...">← Back</button>
```

This matches the accessibility bar set in GAME-008.

game/web/src/main.ts
cgruber: Fixed. Changed `campaignParam ? ... : ...` to `activeCampaign ? ... : ...` in the `renderScenarioCards` card click handler so that unknown or missing `?campaign=` values never leak into card navigation URLs — only the validated `activeCampaign` value does.

game/web/src/main.ts
cgruber: Fixed. Applied the same change in `showWipWarning`'s `onConfirm` handler — `campaignParam` replaced with `activeCampaign` so WIP confirmations also use the validated campaign value rather than the raw query param.

game/web/e2e/scenarios.spec.ts
cgruber: Fixed. Added an e2e test: "unknown ?campaign= falls back to all scenarios with Back button hidden". It navigates to `/?campaign=unknown-campaign`, asserts the scenario select screen is shown, verifies the Back button is not visible, and confirms all scenario cards are present. All 81 tests pass.

game/web/index.html
cgruber: Fixed. Added `aria-label="Back to campaign select"` to the Back button in `index.html`. Screen readers will now announce a consistent, descriptive label rather than the bare arrow character.

game/release.main.kts
```diff
@@ -339,6 +339,19 @@ class Deploy : CliktCommand(
...
+                    indexHtml.readText().replace(
+                        """s.src = "bundle.js";""",
+                        """s.src = "bundle.js?v=$version";"""
+                    )
+                )
+            }
+
             val deployMeta = DeployMetadata(
                 version = version,
                 commit = meta.commitId,
```
cgruber, line 362: The patch uses `String.replace()` which silently no-ops when the target literal isn't found — and the outer `if (indexHtml.exists())` guard only handles a missing file. Two silent failure paths remain:

1. **String not found**: If `index.html` is restructured in a future build (e.g. the loader script is inlined differently), `replace` returns the original text and the file is written back unchanged. The deploy succeeds but cache busting never fires — no error, no log.
2. **File not found**: Already guarded, but silently skips rather than warning. A deploy without the patch is a silent regression.

Suggestion: assert the patch was actually applied, or at least log a warning:

```kotlin
val original = indexHtml.readText()
val patched = original.replace(
    """s.src = "bundle.js";""",
    """s.src = "bundle.js?v=$version";"""
)
if (patched == original) {
    System.err.println("WARNING: cache-bust patch had no effect — bundle.js src literal not found in index.html")
}
indexHtml.writeText(patched)
```

This is intentionally a warning rather than a hard `check()` failure, since deploy should still succeed (just without the cache-bust), but the operator should know.

game/web/e2e/scenarios.spec.ts
```diff
@@ -881,3 +881,45 @@ test("about page: accessible from select screen and shows educational content",
...
+test("campaign select: unknown ?campaign= falls back to all scenarios with Back button hidden", async ({ page }) => {
+  await page.goto("/?campaign=bogus");
+  await expect(page.locator("#scenario-select")).toBeVisible({ timeout: 10_000 });
+  // Unknown campaign → full manifest fallback (9 cards)
+  const cards = page.locator(".scenario-card");
+  const count = await cards.count();
+  expect(count).toBe(9);
+  // Back button must not be visible — bogus param must not leak
+  await expect(page.locator("#btn-back-to-campaign")).not.toBeVisible();
+});
```
cgruber, line 925: Nit: `expect(count).toBe(9)` hardcodes the total scenario count. If a new scenario is added to `SCENARIO_MANIFEST`, this test (and the similar one at line 57–63) will fail with a confusing failure message unrelated to the campaign feature being tested.

Since the intent is a regression guard (unknown campaign must not filter), a `>= 8` bound would be more resilient:

```ts
expect(count).toBeGreaterThanOrEqual(8);
```

Same applies to the "no ?campaign= shows all 9 scenarios" test above. Not blocking, but worth noting for when the scenario set grows.

game/release.main.kts
cgruber: Good catch. Added a warning (stderr) when the substitution has no effect, so operators know cache-busting was skipped and will see a clear message to update `release.main.kts`. The write only happens when the replacement succeeds.

game/web/e2e/scenarios.spec.ts
cgruber: Intentional — these tests are regression guards for the fallback behavior specifically. `toBe(9)` means "the fallback shows *all* scenarios in SCENARIO_MANIFEST," and if a 10th scenario is ever added, these tests *should* fail to prompt a deliberate update. A dynamic `.toBeGreaterThan(0)` would let the fallback silently return zero cards and still pass. Leaving as-is.

Issue #164, opened 2026-05-01
Main menu / title screen


cgruber:
Main menu / title screen

**Ticket:** [GAME-050](https://github.com/cgruber/redistricting-sim/blob/6051f02b7a2809bb102d59de0ab8c5a007ac03c7/thoughts/shared/tickets/GAME-050-main-menu.md)

PR #167, opened 2026-05-01
fix(GAME-050): Continue greyed out when no progress; hide menu in showEditor

cgruber:
## Prerequisites
- N/A (parent PR #165 already merged)

## Summary
- Continue button is now always visible, disabled (greyed out) when no last-played scenario exists — fixes staging feedback requesting "present but greyed out for consistency"
- `showEditor()` now explicitly hides `#main-menu` as a defensive guard — fixes overlay bug where main menu appeared above the game editor after intro dismissal

## Validation performed
- e2e tests updated to reflect new disabled-not-hidden behavior for Continue button
- Full e2e suite passes locally (90 tests)

- game web frontend only

- POST-MERGE: `cd game && ./release.main.kts -- prepare && ./release.main.kts -- deploy --env staging`

## Tickets addressed
- see #164 (GAME-050)

## Rollback
- Revert to parent commit #165

### DIFF
--- a/game/web/e2e/main-menu.spec.ts
+++ b/game/web/e2e/main-menu.spec.ts
@@ -35,16 +35,17 @@ test("main menu: About button opens about page", async ({ page }) => {
   await expect(page.locator("#main-menu")).not.toBeVisible();
 });
 
-test("main menu: Continue absent when no campaign progress exists", async ({ page }) => {
+test("main menu: Continue visible but disabled when no campaign progress exists", async ({ page }) => {
   await page.goto("/");
   // Ensure no last-played key
   await page.evaluate((key) => localStorage.removeItem(key), LAST_PLAYED_KEY);
   await page.reload();
   await expect(page.locator("#main-menu")).toBeVisible({ timeout: 10_000 });
-  await expect(page.locator("#btn-main-continue")).not.toBeVisible();
+  await expect(page.locator("#btn-main-continue")).toBeVisible();
+  await expect(page.locator("#btn-main-continue")).toBeDisabled();
 });
 
-test("main menu: Continue present when lastPlayedScenarioId set in localStorage", async ({ page }) => {
+test("main menu: Continue enabled when lastPlayedScenarioId set in localStorage", async ({ page }) => {
   await page.goto("/");
   await page.evaluate(
     ([key, val]) => localStorage.setItem(key, val),
@@ -53,6 +54,7 @@ test("main menu: Continue present when lastPlayedScenarioId set in localStorage"
   await page.reload();
   await expect(page.locator("#main-menu")).toBeVisible({ timeout: 10_000 });
   await expect(page.locator("#btn-main-continue")).toBeVisible();
+  await expect(page.locator("#btn-main-continue")).toBeEnabled();
 });
 
 test("main menu: Continue navigates directly to last played scenario", async ({ page }) => {
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -263,12 +263,12 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		const continueBtn = document.getElementById("btn-main-continue") as HTMLButtonElement | null;
 		if (continueBtn) {
 			if (lastPlayedId !== null) {
-				continueBtn.hidden = false;
+				continueBtn.disabled = false;
 				continueBtn.addEventListener("click", () => {
 					window.location.assign(buildContinueUrl(lastPlayedId));
 				});
 			} else {
-				continueBtn.hidden = true;
+				continueBtn.disabled = true;
 			}
 		}
 
@@ -502,6 +502,7 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		introController?.abort();
 		introController = null;
 		introScreen?.classList.add("hidden");
+		document.getElementById("main-menu")?.classList.add("hidden");
 		appHeader!.style.display = "";
 		mainEl!.style.display = "";
 	}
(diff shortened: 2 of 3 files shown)


game/web/src/main.ts
```diff
@@ -502,6 +502,7 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		introController?.abort();
 		introController = null;
 		introScreen?.classList.add("hidden");
+		document.getElementById("main-menu")?.classList.add("hidden");
```
cgruber, line 505: The defensive `classList.add("hidden")` on main-menu in `showEditor()` is logically safe, but tracing the routing reveals it's a dead path in the current state machine:

- If a valid `?s=<id>` is present, the IIFE falls through to scenario load → eventually calls `showEditor()`. In this path, `showMainMenu()` is never called, so `#main-menu` still has its initial `class="hidden"` from the HTML — nothing to hide.
- If no `?s=` is present, the IIFE calls `showMainMenu()` (or `showScenarioSelect()`) and **returns**. `showEditor()` is unreachable in this branch within a single page load.
- The Continue button fires `window.location.assign(...)` — a full navigation — so there's no same-page transition from menu to editor.

Two possible interpretations:

1. **There was a real user-visible overlay repro.** If so, this fix doesn't address the actual root cause (given the routing, the bug must live elsewhere — CSS layering, z-index, or a path not traced here). It would be masking the symptom.
2. **This is preemptive defensive symmetry** (mirrors the existing `introScreen?.classList.add("hidden")` just above it). Fine, but the commit/PR description doesn't say so.

**Request**: Can you confirm whether you had a reproducible repro of the overlay, and under what URL/state? If it was preemptive symmetry, a short note in the commit message would make future readers confident it's intentional. If there was a real repro, the root cause may still be open.

Not a blocker — the line is safe either way — but the answer determines whether there's a lurking bug.

game/web/e2e/main-menu.spec.ts
```diff
@@ -35,16 +35,17 @@ test("main menu: About button opens about page", async ({ page }) => {
...
-test("main menu: Continue absent when no campaign progress exists", async ({ page }) => {
+test("main menu: Continue visible but disabled when no campaign progress exists", async ({ page }) => {
   await page.goto("/");
   // Ensure no last-played key
   await page.evaluate((key) => localStorage.removeItem(key), LAST_PLAYED_KEY);
   await page.reload();
   await expect(page.locator("#main-menu")).toBeVisible({ timeout: 10_000 });
-  await expect(page.locator("#btn-main-continue")).not.toBeVisible();
+  await expect(page.locator("#btn-main-continue")).toBeVisible();
+  await expect(page.locator("#btn-main-continue")).toBeDisabled();
```
cgruber, line 45: Minor observation: the test confirms the disabled button is visible and `.toBeDisabled()`, but doesn't assert it doesn't navigate when clicked. Browsers won't fire `click` events on `disabled` buttons, so this is implicitly safe — but if you want belt-and-suspenders coverage, a `page.locator("#btn-main-continue").click({ force: true })` + URL stability assertion would make it explicit. Not a blocker; just worth noting for completeness.

cgruber:
## Critique: GAME-050 main menu bug fix

**Scope**: Clean. All three files are in-scope for the main menu UX fix. No unrelated changes.

### index.html — `hidden` → `disabled` on Continue button

Correct. `hidden` removes the element from the DOM/layout entirely; `disabled` keeps it visible but inert. The sibling `btn-main-load` and `btn-main-settings` already follow the visible-but-disabled convention, so Continue now matches. The change aligns with the stated UX intent (always show Continue, grey it out when no progress).

### main.ts `showMainMenu()` — `hidden = false/true` → `disabled = false/true`

Correct and necessary to match the HTML change. The logic is sound: if `lastPlayedId !== null`, enable and wire the click handler; otherwise disable. The CSS rule `#main-menu-nav button:disabled { color: #555e6e; border-color: #222e40; cursor: default; }` visually distinguishes the disabled state, so the UX is clear.

### main.ts `showEditor()` — defensive main-menu hide

One question raised inline (see thread on line 505): by routing analysis, this line appears to be dead code in the current state machine — `showMainMenu()` and `showEditor()` live in mutually exclusive return branches. The fix is safe either way, but it would help to know whether you had a reproducible repro of the overlay. If yes, the real root cause may still be open. If it's preemptive symmetry (mirrors the `introScreen` hide just above), a note in the commit message would make it intentional.

### e2e tests

Tests correctly reflect the new UX: renamed test descriptions, `toBeVisible()` instead of `not.toBeVisible()`, and `toBeDisabled()`/`toBeEnabled()` assertions. The "Continue navigates to last played scenario" test (line 60) remains unchanged and implicitly validates the enabled path end-to-end. Minor observation left inline on line 45 about absence of a no-navigate assertion on the disabled path — not a blocker.

---

**Overall**: LGTM with one question to resolve (Q3, is there a real repro?). The core `hidden → disabled` change is correct, the tests match the new UX, and the CSS already styles disabled buttons correctly. If the `showEditor` line is preemptive defensive code, land it with a note and this is ready to merge.

game/web/src/main.ts
cgruber: Real repro from staging: after clicking Continue, going through intro slides, then dismissing via the button, the main menu title block appeared in a stripe above the game editor. Static analysis of the state machine didn't surface the path, but the user observed it consistently. The defensive guard matches the symmetry of the adjacent introScreen hide and costs nothing. If we uncover the root cause later we can remove it, but for now it fixes the observable bug.

game/web/e2e/main-menu.spec.ts
cgruber: Agreed — clicking a disabled button is blocked by the browser natively so no navigation test needed. The `toBeDisabled()` assertion covers the inert state.

Issue #168, opened 2026-05-01
Campaign select screen


cgruber:
Campaign select screen

**Ticket:** [GAME-049](https://github.com/cgruber/redistricting-sim/blob/6051f02b7a2809bb102d59de0ab8c5a007ac03c7/thoughts/shared/tickets/GAME-049-campaign-select-screen.md)

PR #169, opened 2026-05-01
feat(GAME-049): campaign select screen

cgruber:
## Prerequisites
- N/A (GAME-050 already merged)

## Summary
- Adds campaign select screen at `?view=campaigns`: shows both campaigns (Tutorial, Educational) with title, description, and live progress indicator ("N / M scenarios complete") pulled from localStorage
- Clicking a campaign card navigates to `?campaign=<id>` (scenario select for that campaign)
- Back button returns to main menu (`./`)
- Replaces the GAME-049 stub that was routing `?view=campaigns` to the scenario select screen

## Validation performed
- 6 new e2e tests all pass locally (96 total tests pass)

- game web frontend only

- POST-MERGE: `cd game && ./release.main.kts -- prepare && ./release.main.kts -- deploy --env staging`

## Tickets addressed
- see #168 (GAME-049)

## Rollback
- Revert to parent commit

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -255,6 +255,47 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		return `./?s=${scenarioId}`;
 	}
 
+	function showCampaignSelect() {
+		const el = document.getElementById("campaign-select");
+		if (!el) return;
+
+		const cardsEl = document.getElementById("campaign-cards");
+		if (cardsEl) {
+			cardsEl.innerHTML = "";
+			for (const campaign of CAMPAIGN_REGISTRY) {
+				const completed = campaign.scenarioIds.filter((id) => isCompleted(progress, id)).length;
+				const total = campaign.scenarioIds.length;
+				const card = document.createElement("div");
+				card.className = "campaign-card";
+				card.setAttribute("role", "button");
+				card.setAttribute("tabindex", "0");
+				card.setAttribute("aria-label", campaign.title);
+
+				const heading = document.createElement("h2");
+				heading.textContent = campaign.title;
+				const desc = document.createElement("p");
+				desc.textContent = campaign.description;
+				const prog = document.createElement("div");
+				prog.className = "campaign-progress";
+				prog.textContent = `${completed} / ${total} scenarios complete`;
+				card.append(heading, desc, prog);
+
+				const navigate = () => window.location.assign(`./?campaign=${campaign.id}`);
+				card.addEventListener("click", navigate);
+				card.addEventListener("keydown", (e: KeyboardEvent) => {
+					if (e.key === "Enter" || e.key === " ") { e.preventDefault(); navigate(); }
+				});
+				cardsEl.appendChild(card);
+			}
+		}
+
+		document.getElementById("btn-campaign-back")?.addEventListener("click", () => {
+			window.location.assign("./");
+		});
+
+		el.classList.remove("hidden");
+	}
+
 	function showMainMenu() {
 		const mainMenuEl = document.getElementById("main-menu");
 		if (!mainMenuEl) return;
@@ -350,7 +391,7 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		if (campaignParam !== "" || view === "scenarios") {
 			showScenarioSelect();
 		} else if (view === "campaigns") {
-			showScenarioSelect(); // GAME-049 stub: campaign select not yet built
+			showCampaignSelect();
 		} else {
 			showMainMenu();
 		}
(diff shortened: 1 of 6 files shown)


game/web/src/main.ts
```diff
@@ -255,6 +255,39 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
+
+		const cardsEl = document.getElementById("campaign-cards");
+		if (cardsEl) {
+			cardsEl.innerHTML = "";
+			for (const campaign of CAMPAIGN_REGISTRY) {
+				const completed = campaign.scenarioIds.filter((id) => isCompleted(progress, id)).length;
+				const total = campaign.scenarioIds.length;
+				const card = document.createElement("div");
+				card.className = "campaign-card";
+				card.setAttribute("role", "button");
```
cgruber, line 270: `.campaign-card` is a `<div role="button">` with `aria-label`, but it has no `tabindex="0"` and no keydown handler for Enter/Space. Keyboard users can't focus or activate these cards despite the `role="button"` promise to the accessibility tree. Screen readers will announce them as buttons, but Tab focus skips over them entirely.

Compare to `.sc-play-btn` in the scenario-select screen, which uses a real `<button>` element and gets all of this for free.

Simplest fix — either replace each `.campaign-card` `<div>` with a `<button>` (the card becomes the button, style accordingly), or add `tabindex="0"` plus a keydown handler:

```ts
card.setAttribute("tabindex", "0");
card.addEventListener("keydown", (e) => {
    if (e.key === "Enter" || e.key === " ") {
        e.preventDefault();
        window.location.assign(`./?campaign=${campaign.id}`);
    }
});
```

GAME-008 just shipped basic a11y — this would be a quiet regression against that baseline.

game/web/src/main.ts
```diff
@@ -255,6 +255,39 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
+				card.setAttribute("aria-label", campaign.title);
+				card.innerHTML = `
```
cgruber: `campaign.title` and `campaign.description` are injected via `innerHTML`. This is safe today because both values come from the static `CAMPAIGN_REGISTRY` in `campaigns.ts` and are compile-time constants.

The risk is precedent: the `innerHTML` sink is now established at this location, and a future contributor adding a dynamically-sourced campaign (e.g., loaded from JSON or user input) could inadvertently flow untrusted content through it without realising the injection path exists.

Low-cost defensive hardening — build the card structure with `createElement`/`textContent` instead:

```ts
const h2 = document.createElement("h2");
h2.textContent = campaign.title;
const p = document.createElement("p");
p.textContent = campaign.description;
const progress = document.createElement("div");
progress.className = "campaign-progress";
progress.textContent = `${completed} / ${total} scenarios complete`;
card.append(h2, p, progress);
```

Not a blocker given the current data source, but worth doing while the code is fresh.

cgruber (commented):
## Review: GAME-049 Campaign Select Screen

**Scope** — all 6 files are in-scope. No stray changes.

**AC coverage** — all 5 ticket AC bullets are covered by the e2e suite:
- Both campaigns visible ✓ (test 1)
- Tutorial navigation ✓ (test 2)
- Educational navigation ✓ (test 3)
- 0 / 2 progress with fresh localStorage ✓ (test 4)
- Back button present ✓ (test 5)
- Back button navigates to main menu ✓ (test 6 — bonus beyond the ticket's listed tests)

**Progress counting** — `campaign.scenarioIds.filter(id => isCompleted(progress, id)).length` is correct. Counts all completions regardless of unlock order, matching the "N / M scenarios complete" wording.

**z-index** — campaign-select (88) sits correctly between scenario-select (85) and main-menu (90). The screens are mutually exclusive via `.hidden`, so stacking order is decorative, but the convention is consistent with existing screens.

**XSS** — `innerHTML` injection is safe today (values are compile-time constants in `CAMPAIGN_REGISTRY`). See inline comment for a defensive suggestion.

**Two inline issues noted:**

1. **Keyboard accessibility** (line 270) — `<div role="button">` without `tabindex="0"` or keydown handler is not keyboard-focusable. This regresses against the GAME-008 a11y baseline. Should be fixed before merge.

2. **innerHTML sink** (line 272) — low-risk defensive suggestion to use `textContent` / `createElement` instead.

Otherwise LGTM. The implementation is clean, the routing change is minimal and correct, and the CSS matches the existing screen conventions.

game/web/src/main.ts
cgruber: Fixed: added `tabindex="0"` and a keydown handler for Enter/Space that calls the same navigate function as the click handler.

game/web/src/main.ts
cgruber: Fixed: switched to createElement + textContent for all three child elements (h2, p, campaign-progress div). innerHTML is gone from card construction.

Issue #171, opened 2026-05-01
GAME-054: Remove legacy scenario select and fix invalid campaign redirect


cgruber:
Remove standalone `?view=scenarios` route and fix `?campaign=<bogus>` redirect to main menu.

Ticket: GAME-054 (`thoughts/shared/tickets/GAME-054-remove-legacy-scenario-select.md`)

PR #172, opened 2026-05-01
feat(GAME-054): remove legacy scenario select + fix invalid campaign redirect

cgruber:
## Prerequisites
- N/A

## Summary
- `?campaign=<bogus>` now redirects to main menu instead of showing all 9 scenarios
- `?view=scenarios` (legacy URL) redirects to main menu; campaigns are now the only path to scenario select
- Unknown `?s=<id>` redirects to main menu instead of showing scenario select
- Locked scenario without campaign context redirects to main menu (in-campaign locked scenario still shows campaign's scenario select)
- `backUrl` for no-campaign context updated from `./?view=scenarios` to `./`; error screen back button label updated accordingly
- All affected e2e tests updated: helpers in sprint1/2/3 go directly to `/?s=tutorial-002`; progression/WIP tests use `?campaign=tutorial` or `?campaign=educational`; routing tests in scenarios.spec.ts inverted to verify new redirect behavior

## Validation performed
- [x] `bazel test //web:app_typecheck_test` passes
- [x] `bazel build //web:app` passes

- pastthepost.gg (game web app — routing behavior change)

- Deploy to staging: `./release.main.kts -- prepare && ./release.main.kts -- deploy --env staging`
- Smoke test: verify `?campaign=bogus` and `?view=scenarios` redirect to main menu
- Deploy to production: `./release.main.kts -- deploy --env production`

## Tickets addressed
- see #171 (GAME-054)

## Rollback
- Revert PR; redeploy previous version tag

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -149,16 +149,21 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 	const urlParams = new URLSearchParams(window.location.search);
 	const campaignParam = (urlParams.get("campaign") ?? "").replace(/[^a-z0-9-]/g, "");
 	const activeCampaign = campaignParam !== "" ? getCampaign(campaignParam) : undefined;
+	// Unknown campaign ID → redirect to main menu rather than falling back to all scenarios.
+	if (campaignParam !== "" && activeCampaign === undefined) {
+		window.location.replace("./");
+		return;
+	}
 	// When a campaign is active, show only that campaign's scenarios in manifest order.
-	// Falls back to the full manifest when no (or unknown) ?campaign= is provided.
 	const activeList: ReadonlyArray<{ id: string; title: string }> = activeCampaign
 		? (activeCampaign.scenarioIds
 				.map((id) => MANIFEST_BY_ID.get(id))
 				.filter((e): e is { id: string; title: string } => e !== undefined))
 		: (SCENARIO_MANIFEST as ReadonlyArray<{ id: string; title: string }>);
 
-	// URL to return to when leaving a scenario — preserves campaign context if present.
-	const backUrl = campaignParam !== "" ? `./?campaign=${campaignParam}` : "./?view=scenarios";
+	// URL and label for returning from a scenario — preserves campaign context if present.
+	const backUrl = campaignParam !== "" ? `./?campaign=${campaignParam}` : "./";
+	const backLabel = campaignParam !== "" ? "← Back to Scenarios" : "← Main Menu";
 
 	// ── Scenario select screen (GAME-018 / GAME-021) ──────────────────────────
 	// Rendered from the static manifest + localStorage progress.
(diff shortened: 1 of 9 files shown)


game/web/src/main.ts
```diff
@@ -374,24 +379,33 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 
 	let entryToLoad: { id: string; title: string };
 	if (requestedId !== "" && requestedEntry === undefined) {
-		// Unknown scenario ID (or not in current campaign) — redirect to select screen
-		showScenarioSelect();
+		// Unknown scenario ID (or not in current campaign) — redirect to main menu
+		window.location.replace("./");
 		return;
 	} else if (requestedEntry !== undefined) {
```
cgruber, line 390: **Asymmetry with locked-scenario branch:** when `?campaign=tutorial&s=garbage` is used (unknown scenario ID in a valid campaign context), the code redirects to `./` (main menu), discarding the campaign context. A few lines lower, the _locked_ scenario case correctly calls `showScenarioSelect()` when `activeCampaign` is defined, preserving that context.

Should the unknown-ID-in-campaign case do the same? If a user bookmarks `?campaign=tutorial&s=typo`, the current code ejects them to the main menu rather than showing them the tutorial campaign's scenario list. The locked branch explicitly handles this distinction — the unknown-id branch should probably mirror it:

```ts
if (requestedId !== "" && requestedEntry === undefined) {
    if (activeCampaign) {
        showScenarioSelect(); // stay in campaign context
    } else {
        window.location.replace("./"); // no campaign — go to main menu
    }
    return;
}
```

If the intent is "unknown ID always goes to main menu regardless of campaign," please add a comment explaining why (e.g., security / no partial-URL guessing).

thoughts/shared/tickets/GAME-054-remove-legacy-scenario-select.md
```diff
@@ -0,0 +1,46 @@
...
+## Goals / Acceptance Criteria
+
+- [ ] `?campaign=<invalid>` (any unrecognized id) → `window.location.replace("./")` → main menu
+- [ ] `?view=scenarios` → `window.location.replace("./")` → main menu
+- [ ] Unknown `?s=<id>` → `window.location.replace("./")` → main menu (was: scenario select)
+- [ ] Locked scenario (no campaign context) → main menu (was: scenario select)
+- [ ] `backUrl` for no-campaign scenarios points to `./` (main menu), not `./?view=scenarios`
+- [ ] Error screen back button label updated to "← Main Menu" for no-campaign context
+
+## Test Coverage
```
cgruber, line 32: The ticket lists this as an acceptance criterion:

> Unknown `?s=<id>` → `window.location.replace("./")` → main menu

But no e2e test was added for it. The existing "lock gate" test covers the _locked_ case; there's no test covering `?s=garbage` → `#main-menu`. Minor omission given the other thorough coverage, but worth a follow-up or a quick test added here before merge.

thoughts/shared/tickets/GAME-054-remove-legacy-scenario-select.md
```diff
@@ -0,0 +1,46 @@
...
+
+- [ ] e2e: `?campaign=bogus` → `#main-menu` visible, `#scenario-select` not visible
+- [ ] e2e: `?view=scenarios` → `#main-menu` visible, `#scenario-select` not visible
+- [ ] e2e: All existing scenario-select-dependent tests updated to use campaign URLs
+- [ ] e2e: Lock gate test updated to expect `#main-menu` not `#scenario-select`
+- [ ] e2e: Debug force-win test updated (no-campaign → `backUrl` → main menu)
+
+## References
```
cgruber, line 41: GAME-051 (`GAME-051-ingame-navigation-cleanup.md`) still describes the "Return to Scenarios" option as navigating to `?view=scenarios` when there's no campaign context (line 30–31: `?view=scenarios` if no campaign context). Now that this PR removes `?view=scenarios` as a functional route (it redirects to main menu), that acceptance criterion is contradicted — "Return to Scenarios" without campaign context will just loop back to main menu rather than reaching a scenario list.

GAME-051 will need its AC updated before implementation: either "Return to Scenarios" always requires campaign context (and is hidden/disabled without one), or it has a different no-campaign fallback. Worth updating the ticket before someone picks it up and implements against a now-wrong spec.

game/web/src/main.ts
cgruber: Fixed: unknown ?s= with an active campaign now calls showScenarioSelect() instead of window.location.replace("./"), matching the locked-scenario branch. No-campaign unknown-ID still redirects to main menu.

thoughts/shared/tickets/GAME-054-remove-legacy-scenario-select.md
cgruber: Fixed: added e2e test "routing: unknown ?s= without campaign redirects to main menu" to scenarios.spec.ts. Also updated the ticket AC to distinguish the two cases (with campaign vs without).

thoughts/shared/tickets/GAME-054-remove-legacy-scenario-select.md
cgruber: Fixed: updated GAME-051 AC to reflect that standalone scenario select is removed. "Return to Scenarios" now only appears in campaign context; no-campaign in-game view gets only "Return to Main Menu".

Issue #176, opened 2026-05-01
GAME-008: Accessibility pass — color blindness, keyboard nav, screen reader support


cgruber:
Full a11y pass: color-blind-safe palettes (Okabe-Ito for districts, PuOr for lean view), keyboard precinct navigation (arrow-key + number-key), focus rings, prefers-reduced-motion, and automated e2e ARIA tests. WCAG 2.1 AA target. See ticket: thoughts/shared/tickets/GAME-008-accessibility.md

cgruber:
All AC met and merged in PR #177 (v0.0.13). Deployed to staging.

PR #177, opened 2026-05-01
feat(GAME-008): full accessibility pass — CVD-safe palettes, keyboard nav, focus rings, a11y tests

cgruber:
## Summary

Full WCAG 2.1 AA accessibility pass for GAME-008. Sprint 9 (PR #129) already added basic ARIA labels and aria-live regions; this PR completes the remaining AC: color-blind-safe palettes, keyboard precinct navigation, focus rings, and automated a11y e2e tests.

- **Color-blind-safe district palette** — replaced the old tableau-style 5 colors with the Okabe-Ito (2008) CVD-safe set (amber / sky-blue / bluish-green / mauve / vermilion); safe for deuteranopia, protanopia, and tritanopia
- **Lean view palette** — replaced `d3.interpolateRdBu` with `d3.interpolatePuOr` (purple-orange); avoids semantic collision with party colors (D=blue, R=red); clamped to [0.1, 0.9] for dark-background contrast
- **Party colors aligned with lean view** — D→purple (`#7b35a8`), R→orange (`#c96d00`); vote-bar CSS gradient updated to match; winner badges now use consistent color language across the whole UI
- **Keyboard precinct navigation** — SVG changed to `role="application"` + `tabindex="0"`; arrow keys navigate hex neighbors, number keys 1–5 assign district, Space assigns to active district; yellow dashed focus ring (#F0E442, Okabe-Ito); `aria-label` updates dynamically with focused precinct + district context
- **Focus rings** — `:focus-visible` global rule (2px sky-blue outline); district buttons use white outline to match active-border convention; SVG container gets 4px offset ring
- **`prefers-reduced-motion`** — media query collapses all transitions/animations to 0.01ms
- **6 new e2e a11y tests** (`e2e/a11y.spec.ts`): HTML lang, main-menu ARIA roles, game-screen ARIA labels, tab order, SVG keyboard nav, keyboard district assignment; 109 total tests pass

- `game/web/` — TypeScript source, HTML, CSS, e2e tests
- No server-side changes; static browser app only

## Validation performed

- `bazel build //web:app_typecheck_test` — passes (no type errors)
- `bazel test //web:e2e_test` — 109/109 pass locally
- Deployed to dev (`vTEST-6474090261ca`) and manually verified:
  - District colors visually distinct and game-y on dark background
  - Lean view shows purple-orange gradient (not red-blue)
  - Party winner badges show orange/purple consistently with lean view
  - Keyboard nav: click SVG → ArrowDown focuses first precinct (yellow dashed ring) → number keys assign
  - Tab key cycles through all enabled controls with visible focus ring
  - VoiceOver test: manual — SVG aria-label updates correctly on keyboard nav (noted in PR, not automated)

## Deployment steps

Standard dev→staging→prod pipeline. Staging deploy follows merge.

## Tickets addressed

- see #176 (GAME-008 full accessibility pass)
- Also files GAME-055 (party names from scenario data) and GAME-056 (playtest feedback intake) as follow-on tickets — not blocked on this PR

## Rollback

Revert this commit. No data migrations; no server-side changes. District colors and lean palette revert to previous values.

### DIFF
--- a/game/web/e2e/a11y.spec.ts
+++ b/game/web/e2e/a11y.spec.ts
@@ -1,0 +1,142 @@
+import { test, expect } from "@playwright/test";
+
+/**
+ * GAME-008 Accessibility e2e tests.
+ *
+ * Targeted ARIA and keyboard assertions — no axe-core dependency.
+ * Covers:
+ *   1. HTML lang attribute
+ *   2. Main menu ARIA roles
+ *   3. Game screen ARIA labels
+ *   4. Control button tab order
+ *   5. SVG keyboard navigation
+ *   6. Keyboard district assignment
+ */
+
+// tutorial-001 is campaign-only; must include campaign param to bypass routing guard.
+const SCENARIO_URL = "/?campaign=tutorial&s=tutorial-001&debug=true";
+
+/** Navigate to the game editor, skip the intro, and wait for the hex grid. */
+async function loadEditor(page: import("@playwright/test").Page): Promise<void> {
+  await page.goto(SCENARIO_URL);
+  const skip = page.locator("#btn-intro-skip");
+  await expect(skip).toBeVisible({ timeout: 10_000 });
+  await skip.click();
+  await expect(page.locator("#map-svg path.hex").first()).toBeVisible({ timeout: 10_000 });
+}
+
+test.describe("GAME-008 Accessibility", () => {
+  // ─── Test 1: HTML lang attribute ──────────────────────────────────────────
+
+  test("html element has lang attribute", async ({ page }) => {
+    await page.goto("/");
+    const lang = await page.evaluate(() => document.documentElement.lang);
+    expect(lang).toBeTruthy();
+  });
+
+  // ─── Test 2: Main menu ARIA roles ─────────────────────────────────────────
+
+  test("main menu has correct ARIA roles and focusable buttons", async ({ page }) => {
+    await page.goto("/");
+    await expect(page.locator("#main-menu")).toBeVisible({ timeout: 10_000 });
+
+    // #main-menu must carry role="main"
+    const menuRole = await page.locator("#main-menu").getAttribute("role");
+    expect(menuRole).toBe("main");
+
+    // #main-menu-nav must declare its navigation purpose
+    const navRole = await page.locator("#main-menu-nav").getAttribute("role");
+    const navLabel = await page.locator("#main-menu-nav").getAttribute("aria-label");
+    expect(navRole === "navigation" || (navLabel !== null && navLabel.length > 0)).toBeTruthy();
+
+    // New Campaign button must be focusable
+    await page.locator("#btn-main-new-campaign").focus();
+    const focused = await page.evaluate(() => document.activeElement?.id);
+    expect(focused).toBe("btn-main-new-campaign");
+  });
+
+  // ─── Test 3: Game screen ARIA labels ──────────────────────────────────────
+
+  test("game screen elements have required ARIA labels and roles", async ({ page }) => {
+    await loadEditor(page);
+
+    // SVG map must be keyboard-accessible and labelled
+    const svg = page.locator("#map-svg");
+    const svgRole = await svg.getAttribute("role");
+    expect(svgRole).toBe("application");
+
+    const svgTabindex = await svg.getAttribute("tabindex");
+    expect(svgTabindex).toBe("0");
+
+    const svgAriaLabel = await svg.getAttribute("aria-label");
+    expect(svgAriaLabel).toBeTruthy();
+
+    // Results container must announce updates to assistive tech
+    const resultsLive = await page.locator("#results-container").getAttribute("aria-live");
+    expect(resultsLive).toBe("polite");
+
+    // All control buttons must have accessible labels
+    for (const id of ["btn-undo", "btn-redo", "btn-view-toggle", "btn-submit"]) {
+      const label = await page.locator(`#${id}`).getAttribute("aria-label");
+      expect(label, `#${id} must have a non-empty aria-label`).toBeTruthy();
+    }
+
+    // District button group must expose its group role
+    const groupRole = await page.locator("#district-buttons").getAttribute("role");
+    expect(groupRole).toBe("group");
+  });
+
+  // ─── Test 4: Control button tab order ─────────────────────────────────────
+
+  test("control buttons appear in tab sequence", async ({ page }) => {
+    await loadEditor(page);
+
+    const focused: string[] = [];
+    for (let i = 0; i < 20; i++) {
+      await page.keyboard.press("Tab");
+      const id = await page.evaluate(() => document.activeElement?.id ?? "");
+      focused.push(id);
+      if (focused.length > 15 && focused.includes("btn-submit")) break;
+    }
+
+    // Only assert always-enabled buttons — disabled buttons (undo, redo, submit)
+    // are correctly excluded from tab order until actions enable them.
+    for (const id of ["btn-view-toggle", "btn-county-toggle", "btn-reset", "map-svg"]) {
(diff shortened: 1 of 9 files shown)


game/web/src/render/mapRenderer.ts
```diff
@@ -629,5 +637,119 @@ export class SvgMapRenderer implements MapRenderer {
...
+					}
+				}
+				return;
+			}
+
+			// Number keys 1–5: assign focused precinct to that district
+			const num = parseInt(e.key, 10);
+			if (num >= 1 && num <= districtCount) {
+				e.preventDefault();
+				this.paintStroke([this.focusedPrecinctId], num);
```
cgruber, line 695: **Bug: aria-label not refreshed after keyboard district assignment**

After `paintStroke` is called here (and in the Space handler below), the SVG's `aria-label` still reflects the *pre-assignment* district state — e.g. "focused: North Ridge, unassigned" — even though the precinct was just assigned. A screen reader user pressing 1–5 or Space gets no audio confirmation that the assignment took effect.

Fix: re-call `setKeyboardFocus(this.focusedPrecinctId, precincts)` immediately after each `paintStroke` invocation. Because `setKeyboardFocus` re-reads `this.getState().assignments`, the label will reflect the new assignment.

```ts
this.paintStroke([this.focusedPrecinctId], num);
this.setKeyboardFocus(this.focusedPrecinctId, precincts); // refresh label post-assignment
```

game/web/src/render/mapRenderer.ts
```diff
@@ -629,5 +637,119 @@ export class SvgMapRenderer implements MapRenderer {
...
+			}
+		});
+
+		// Clear keyboard focus when SVG loses focus
+		svgNode.addEventListener("blur", () => {
+			this.clearKeyboardFocus();
+		});
+	}
+
+	private setKeyboardFocus(precinctId: number, precincts: Precinct[]) {
```
cgruber, line 716: **Bug: mouse hover overwrites keyboard focus ring**

`clearHover()` (existing code, not in this diff) unconditionally sets `stroke: "none"` on `this.hoveredPath`. If the user's mouse passes over the keyboard-focused hex, `initHoverEvents` → `clearHover` will erase the yellow dashed focus ring, leaving no visual indication of which precinct is focused.

The fix is to guard `clearHover` so it skips restoring `stroke: "none"` when the path being cleared is `this.keyboardFocusPath`:

```ts
private clearHover() {
    if (this.hoveredPath === null) return;
    const path = this.hoveredPath;
    this.hoveredPath = null;
    const isKeyboardFocused = path === this.keyboardFocusPath;
    if (!isKeyboardFocused) {
        // only restore stroke to none if not keyboard-focused
        d3.select(path).attr("stroke", "none").attr("stroke-width", 0.5)...
    }
    // restore opacity either way
}
```

A symmetric guard is needed in the hover entry path in `initHoverEvents`: if the hovered hex equals `this.keyboardFocusPath`, skip writing the white hover stroke so the focus ring remains visible.

game/web/src/render/mapRenderer.ts
```diff
@@ -629,5 +637,119 @@ export class SvgMapRenderer implements MapRenderer {
...
+		const path = this.hexGroup
+			.select<SVGPathElement>(`path.hex[data-precinct-id="${precinctId}"]`)
+			.node();
+		if (path === null) return;
+		this.keyboardFocusPath = path;
+
+		// Yellow dashed focus ring — distinct from hover (white) and district fills
+		d3.select(path)
+			.attr("stroke", "#F0E442")
+			.attr("stroke-width", 2 / this.currentK)
```
cgruber, line 730: **Minor: focus ring stroke-width/dasharray stale after zoom**

`stroke-width` and `stroke-dasharray` are set once here using `this.currentK` at the moment focus is set. The `onZoom` handler (in `initZoom`) updates boundary and preview stroke widths as `currentK` changes, but it does not update the keyboard focus ring. If the user focuses a precinct then zooms, the dashed focus ring will appear thinner or thicker than intended relative to the hex size.

The `onZoom` handler should also rescale the focus ring when one is active:

```ts
.on("zoom", (event) => {
    this.currentK = event.transform.k;
    // ... existing stroke updates ...
    if (this.keyboardFocusPath !== null) {
        d3.select(this.keyboardFocusPath)
            .attr("stroke-width", 2 / this.currentK)
            .attr("stroke-dasharray", `${4 / this.currentK},${2 / this.currentK}`);
    }
});
```

Lower priority than the aria-label staleness and hover collision issues, but worth fixing for visual consistency.

game/web/e2e/a11y.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    // Click SVG to focus it, then press ArrowDown to select first precinct
+    await page.locator("#map-svg").click();
+    await page.keyboard.press("ArrowDown");
+
+    const label = await page.locator("#map-svg").getAttribute("aria-label");
+    expect(label).toContain("focused:");
+  });
+
+  // ─── Test 6: Keyboard district assignment ─────────────────────────────────
```
cgruber, line 122: **Missing assertion: `#map-svg` should appear in tab sequence**

The SVG now has `tabindex="0"` (new in this PR), making it a tab stop. Test 4 loops through up to 20 Tab presses and collects focused element IDs, but the assertion only checks `btn-view-toggle`, `btn-county-toggle`, and `btn-reset`. The SVG itself (`#map-svg`) is not asserted.

If `#map-svg` were accidentally removed from the tab order, this test would still pass. Add it to the assertion set:

```ts
for (const id of ["btn-view-toggle", "btn-county-toggle", "btn-reset", "map-svg"]) {
    expect(focused, `#${id} must appear in tab sequence`).toContain(id);
}
```

(Note: `document.activeElement?.id` returns `"map-svg"` for the SVG, so no selector change is needed — just add it to the array.)

game/web/src/render/mapRenderer.ts
cgruber: Fixed. After `paintStroke(…)` in both the number-key and space branches, the handler now calls:

```typescript
const { precincts } = this.getState();
this.setKeyboardFocus(this.focusedPrecinctId, precincts);
```

This re-reads the updated store after the assignment is committed and rebuilds the aria-label with the new district assignment.

game/web/src/render/mapRenderer.ts
cgruber: Fixed with two guards:

1. In `clearHover()`, before restoring stroke, skip paths that are the keyboard focus path:
```typescript
if (this.hoveredPath === this.keyboardFocusPath) {
    this.hoveredPath = null;
    return;
}
```

2. In the `mousemove` handler, after `clearHover()`, skip applying the white hover stroke if the newly hovered path is the focused path:
```typescript
if (path === this.keyboardFocusPath) return;
```

Together these prevent the hover logic from overwriting or erasing the yellow dashed focus ring.

game/web/src/render/mapRenderer.ts
cgruber: Fixed. At the end of the `on("zoom", …)` handler (after the preview-boundary stroke-width update), the focus ring is now rescaled:

```typescript
if (this.keyboardFocusPath !== null) {
    d3.select(this.keyboardFocusPath)
        .attr("stroke-width", 2 / this.currentK)
        .attr("stroke-dasharray", `${4 / this.currentK},${2 / this.currentK}`);
}
```

`this.currentK` is already updated at the top of the handler, so the focus ring scales consistently with boundaries.

game/web/e2e/a11y.spec.ts
cgruber: Fixed. Added `"map-svg"` to the assertion list in test 4:

```typescript
for (const id of ["btn-view-toggle", "btn-county-toggle", "btn-reset", "map-svg"]) {
```

Since `#map-svg` has `tabindex="0"`, it must appear in the tab sequence alongside the control buttons.

Issue #179, opened 2026-05-02
GAME-055: Party names should come from scenario data, not hardcoded labels


cgruber:
Results panel shows 'Red Party'/'Blue Party' instead of scenario-specific names (Ken/Ryu, Cat/Dog). See ticket: thoughts/shared/tickets/GAME-055-party-names-scenario-driven.md

PR #180, opened 2026-05-02
feat(GAME-055): scenario-driven party names in results panel

cgruber:
## Summary

Party names in the results panel now come from scenario data instead of hardcoded `PARTY_LABELS`. Players see "Ken Party wins" and "Ryu Party 52% · Ken Party 48%" in tutorial-001, "Cat Party / Dog Party" in scenario-009, etc. Falls back to "Red Party" / "Blue Party" for any scenario that omits party names.

- `renderResults()` in `panels.ts` accepts an optional `partyLabels` parameter; merges with `PARTY_LABELS` defaults so the fallback is always clean
- `main.ts` derives `partyLabels` from `scenario.parties[0]` → R and `scenario.parties[1]` → D after load, and passes it to `renderResults()`
- Vote-details line (`"Blue X% · Red Y%"`) now uses scenario names
- Result screen (`showResultScreen`) confirmed clean — does not reference party labels
- 2 new e2e tests: tutorial-001 shows Ken/Ryu, scenario-009 shows Cat/Dog; both assert absence of hardcoded "Red Party"/"Blue Party" strings

`game/web/` — TypeScript source and e2e tests only. No server-side changes.

## Validation performed

- `bazel build //web:app_typecheck_test` passes
- `bazel test //web:e2e_test` — 111/111 pass locally

## Deployment steps

Standard pipeline — staging deploy follows merge.

## Tickets addressed

- see #179 (GAME-055)

## Rollback

Revert this commit. No data migrations; no server-side changes.

### DIFF
--- a/game/web/e2e/a11y.spec.ts
+++ b/game/web/e2e/a11y.spec.ts
@@ -140,3 +140,81 @@ test.describe("GAME-008 Accessibility", () => {
     expect(jsErrors).toHaveLength(0);
   });
 });
+
+// ─── GAME-055: Party names ────────────────────────────────────────────────────
+
+test.describe("GAME-055 Party names", () => {
+  /**
+   * Test 1: tutorial-001 results panel shows scenario party names (Ken Party / Ryu Party)
+   * not the generic PARTY_LABELS fallbacks ("Party 1" / "Party 2").
+   *
+   * tutorial-001 starts with all precincts unassigned, so we must paint them first
+   * via window.__gameStore to trigger a non-empty simulationResult.
+   */
+  test("tutorial-001 results panel shows scenario party names not hardcoded labels", async ({ page }) => {
+    await page.goto("/?campaign=tutorial&s=tutorial-001&debug=true");
+    const skip = page.locator("#btn-intro-skip");
+    await expect(skip).toBeVisible({ timeout: 10_000 });
+    await skip.click();
+    await expect(page.locator("#map-svg path.hex").first()).toBeVisible({ timeout: 10_000 });
+
+    // Assign all precincts to districts 1 and 2 (split roughly in half) via the
+    // exposed game store so renderResults() runs with non-empty districtResults.
+    await page.evaluate(() => {
+      const store = (window as unknown as Record<string, {
+        getState: () => {
+          paintStroke: (ids: number[], district: number) => void;
+          precincts: unknown[];
+        };
+      }>)["__gameStore"];
+      if (!store) throw new Error("__gameStore not exposed");
+      const { precincts, paintStroke } = store.getState();
+      const total = precincts.length;
+      const half = Math.floor(total / 2);
+      // District 1 gets first half, district 2 gets second half
+      paintStroke(Array.from({ length: half }, (_, i) => i), 1);
+      paintStroke(Array.from({ length: total - half }, (_, i) => i + half), 2);
+    });
+
+    // Wait for at least one result-district row to appear
+    await expect(page.locator("#results-container .result-district").first()).toBeVisible({ timeout: 5_000 });
+
+    const resultsText = await page.locator("#results-container").innerText();
+
+    // Must NOT contain the generic PARTY_LABELS fallbacks
+    expect(resultsText).not.toContain("Party 1");
+    expect(resultsText).not.toContain("Party 2");
+
+    // Must contain both scenario-specific party names
+    expect(resultsText).toContain("Ken Party");
+    expect(resultsText).toContain("Ryu Party");
+  });
+
+  /**
+   * Test 2: scenario-009 results panel shows Cat Party / Dog Party names.
+   *
+   * scenario-009 ships with all precincts pre-assigned, so results are already
+   * populated on initial load — no painting required.
+   */
+  test("scenario-009 results panel shows scenario party names not hardcoded labels", async ({ page }) => {
+    // scenario-009 is in the educational campaign, accessible via ?s= + ?debug
+    await page.goto("/?campaign=educational&s=scenario-009&debug=true");
+    const skip = page.locator("#btn-intro-skip");
+    await expect(skip).toBeVisible({ timeout: 10_000 });
+    await skip.click();
+    await expect(page.locator("#map-svg path.hex").first()).toBeVisible({ timeout: 10_000 });
+
+    // All precincts are pre-assigned in scenario-009 — results render on page load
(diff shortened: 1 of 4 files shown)


game/web/src/main.ts
```diff
@@ -527,14 +527,21 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		partyIdToKey.set(p.id, SPIKE_PARTY_KEYS[i] ?? "I");
 	});
 
+	// ── Party labels derived from scenario data (GAME-055) ───────────────────
+	// Override the hardcoded PARTY_LABELS fallbacks with scenario party names.
+	const partyLabels: Partial<Record<"R" | "D" | "L" | "G" | "I", string>> = {
+		...(scenario.parties[0]?.name !== undefined ? { R: scenario.parties[0].name } : {}),
+		...(scenario.parties[1]?.name !== undefined ? { D: scenario.parties[1].name } : {}),
```
cgruber: The `partyLabels` object only picks up `parties[0]` → `R` and `parties[1]` → `D`, but `partyIdToKey` just above maps up to five parties using `SPIKE_PARTY_KEYS = ["R", "D", "L", "G", "I"]`. If a future scenario adds a third party, that party will win districts and render as the `PARTY_LABELS` fallback ("Libertarian") in the winner badge rather than its scenario name, silently diverging from the key that was actually assigned.

The fix is to drive `partyLabels` from the same loop already used for `partyIdToKey`:

```ts
const partyLabels: Partial<Record<"R" | "D" | "L" | "G" | "I", string>> = {};
scenario.parties.forEach((p, i) => {
    const key = SPIKE_PARTY_KEYS[i] ?? "I";
    if (p.name !== undefined) partyLabels[key] = p.name;
});
```

This keeps the two mappings in sync regardless of how many parties a scenario defines.

game/web/src/render/panels.ts
```diff
@@ -25,7 +30,7 @@ export function renderResults(container: HTMLElement, state: GameStore): void {
         <div class="winner-badge" style="background:${winnerColor};color:#fff">${winnerLabel} +${marginPct}%</div>
         <div class="vote-bar" style="--d-pct:${dPct}%"></div>
         <div class="vote-details">
-          Blue ${dPct}% · Red ${rPct}% · ${r.precinctCount} precincts · pop ${r.population.toLocaleString()}
+          ${labels.D} ${dPct}% · ${labels.R} ${rPct}% · ${r.precinctCount} precincts · pop ${r.population.toLocaleString()}
```
cgruber, line 33: The vote-details line unconditionally formats `labels.D` and `labels.R`. `voteTotals: PartyShare` carries all five party keys (R, D, L, G, I), so for any scenario where the first party maps to a key other than R or D the wrong share would appear under the correct label, or a zero share appears for a key that carries no votes.

This is a pre-existing 2-party assumption that the PR inherits rather than introduces, so it doesn't need to block merge. But it's worth a note: if a >2-party scenario is ever added, the vote-details line will need to iterate `Object.entries(r.voteTotals)` filtered to non-zero shares rather than hardcoding `.D` and `.R`.

game/web/e2e/a11y.spec.ts
```diff
@@ -140,3 +140,81 @@ test.describe("GAME-008 Accessibility", () => {
...
+    await expect(page.locator("#results-container .result-district").first()).toBeVisible({ timeout: 5_000 });
+
+    const resultsText = await page.locator("#results-container").innerText();
+
+    // Must NOT contain the hardcoded PARTY_LABELS fallbacks
+    expect(resultsText).not.toContain("Red Party");
+    expect(resultsText).not.toContain("Blue Party");
+
+    // Must contain at least one of the scenario-specific party names
+    const hasScenarioNames = resultsText.includes("Ken Party") || resultsText.includes("Ryu Party");
```
cgruber: Both `hasScenarioNames` assertions use OR rather than AND:

```ts
const hasScenarioNames = resultsText.includes("Ken Party") || resultsText.includes("Ryu Party");
```

Because the vote-details line now renders `${labels.D} … ${labels.R} …` for every district, both party names will appear in the results text whenever any district renders. The OR allows the test to pass even if only one name appears — e.g. if the label substitution silently breaks for one party. Change both to AND:

```ts
expect(resultsText).toContain("Ken Party");
expect(resultsText).toContain("Ryu Party");
```

game/web/src/main.ts
cgruber: Fixed. Replaced the hardcoded `parties[0]`/`parties[1]` spread with a loop over `partyIdToKey` so all mapped parties (R/D/L/G/I) get their names populated from scenario data. The new code:

```typescript
const partyLabels: Partial<Record<"R" | "D" | "L" | "G" | "I", string>> = {};
scenario.parties.forEach((p) => {
    const key = partyIdToKey.get(p.id);
    if (key !== undefined) partyLabels[key as "R" | "D" | "L" | "G" | "I"] = p.name;
});
```

game/web/src/render/panels.ts
cgruber: No change — pre-existing 2-party assumption in `vote-details` formatting, out of scope for GAME-055.

game/web/e2e/a11y.spec.ts
cgruber: Fixed. Replaced the OR-based `hasScenarioNames` pattern with individual `toContain` assertions for both party names in both tests. E2e suite passes locally (`//web:e2e_test PASSED in 11.3s`).

game/web/src/main.ts
cgruber: Fixed — the `forEach` over `partyIdToKey` correctly handles all scenario parties; `partyIdToKey` is built from the full `scenario.parties` array with index mapping, so any party with a registered key gets its label from the scenario. The critique was based on the prior version; the current implementation already resolves this.

game/web/src/render/panels.ts
cgruber: No change for GAME-055 scope. The vote-details line intentionally shows only the two main parties (R and D) that carry meaningful vote shares in all current scenarios. Displaying all five PartyKey values would add three near-zero lines (L, G, I) in every result row. A full multi-party display can be tracked separately if scenario data ever needs it. This ticket is specifically about names, not the display structure.

game/web/e2e/a11y.spec.ts
cgruber: Fixed — assertions are now two separate `toContain()` calls, not an OR. Each party name is independently required. Lines 189-190 and 217-218 each assert the two party names individually.

PR #192, opened 2026-05-02
feat(GAME-063): asset pipeline — directory structure, Bazel integration, deployable inclusion

cgruber:
## Summary

- Creates `game/web/assets/characters/` and `game/web/assets/audio/` directories with `.gitkeep` placeholders, establishing the directory structure for GAME-060 (character sprites) and GAME-061 (audio clips) to populate.
- Declares a `asset_files` Bazel filegroup in `game/web/BUILD.bazel` covering `assets/**/*` (with `allow_empty = True` so builds pass before real assets land).
- Extends `//web:deployable` genrule to include `asset_files` in srcs and copy files into `assets/characters/` and `assets/audio/` subdirectories within the zip (path-based dispatch, not extension-based).
- Adds `:asset_files` to `//web:e2e_test` data deps so assets are available in runfiles during tests.
- Updates `e2e_test.sh` to copy the `assets/` tree into `SERVE_DIR` (mirrors the existing `scenarios/` pattern).
- Updates `game/serve.sh` dev server to copy `web/assets/` into `_serve_dist/assets/` at startup.

- Local dev server (`game/serve.sh`): now copies assets on startup.
- CI (`//web:deployable`, `//web:e2e_test`): both targets rebuilt with asset support.
- Deployable zip: now includes `assets/characters/` and `assets/audio/` subtrees.

## Validation performed

- `bazel build //web:deployable` succeeds (974 actions, no errors).
- Build output confirms `bazel-bin/web/deployable.zip` produced with `assets/characters/.gitkeep` and `assets/audio/.gitkeep` included (filegroup srcs dispatched by path pattern).
- Reviewed `BUILD.bazel`, `e2e_test.sh`, and `serve.sh` diffs for correctness.

## Deployment steps

No deployment action required. This is infrastructure scaffolding only; no real asset files are included yet. Future GAME-060 and GAME-061 PRs will add the actual SVG/audio files.

## Tickets addressed

- GAME-063: Asset pipeline — directory structure, Bazel integration, deployable inclusion

## Rollback

Revert this PR. No data migrations or runtime state changes; rollback is safe at any time.

### DIFF
--- a/game/web/BUILD.bazel
+++ b/game/web/BUILD.bazel
@@ -82,6 +82,21 @@ filegroup(
     visibility = ["//visibility:public"],
 )
 
+# Static asset files (character SVGs, audio clips) served under /assets/.
+# Populated incrementally by GAME-060 (sprites) and GAME-061 (audio).
+#
+# NOTE: asset_files globs assets/**/* and the deployable genrule uses
+# cp -r to copy the whole assets tree, so new subdirectories are picked up
+# automatically — no changes to BUILD.bazel required when adding new subdirs.
+filegroup(
+    name = "asset_files",
+    srcs = glob(
+        ["assets/**/*"],
+        allow_empty = True,
+    ),
+    visibility = ["//visibility:public"],
+)
+
 # Playwright behavioral (e2e) smoke test — CI-002 Phase 1.
 #
 # Assembles Bazel-built artifacts into a temp dir, starts python3 http.server
@@ -99,6 +114,7 @@ sh_test(
     data = [
         ":app",
         ":html",
+        ":asset_files",
         "//rust:wasm_calc_bindgen",
         "//scenarios:scenario_files",
         "playwright.config.ts",
@@ -135,6 +151,7 @@ genrule(
     srcs = [
         ":app",
         ":html",
+        ":asset_files",
         "//rust:wasm_calc_bindgen",
         "//scenarios:scenario_files",
     ],
@@ -146,7 +163,9 @@ genrule(
         STAGING="$$(mktemp -d)"
         trap "rm -rf $$STAGING" EXIT
 
-        # Iterate over all source files and copy to appropriate location
+        # Iterate over all source files and copy to appropriate location.
+        # Assets are handled separately below via cp -r to preserve tree structure.
+        ASSETS_SRC=""
         for f in $(SRCS); do
             fname="$$(basename "$$f")"
             case "$$fname" in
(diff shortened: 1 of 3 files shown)


game/web/BUILD.bazel
```diff
@@ -156,6 +169,10 @@ genrule(
                 wasm_calc_bindgen.js) cp "$$f" "$$STAGING/" ;;
                 *.json) mkdir -p "$$STAGING/scenarios" && cp "$$f" "$$STAGING/scenarios/" ;;
             esac
+            case "$$f" in
+                */assets/characters/*) mkdir -p "$$STAGING/assets/characters" && cp "$$f" "$$STAGING/assets/characters/" ;;
```
cgruber: **Genrule flattens subdirectory structure; dev/test diverges from deployable**

`serve.sh` and `e2e_test.sh` both use:
```sh
cp -r "${ASSETS_DIR}" "${SERVE_DIR}/assets"
```
This preserves the full tree under `assets/`, so `assets/audio/sfx/click.mp3` arrives at the same path in dev and e2e.

The genrule does:
```sh
*/assets/characters/*) mkdir -p "$$STAGING/assets/characters" && cp "$$f" "$$STAGING/assets/characters/" ;;
*/assets/audio/*)      mkdir -p "$$STAGING/assets/audio"      && cp "$$f" "$$STAGING/assets/audio/" ;;
```
`cp "$$f" "$$STAGING/assets/audio/"` copies the file by basename into the flat target dir. Any nested structure (e.g. `assets/audio/sfx/click.mp3`) silently becomes `assets/audio/click.mp3` in the zip. A URL that works in dev/e2e returns 404 in the deployed build.

**Fix:** replace `cp "$$f" "$$STAGING/assets/..."` with a relative-path-preserving copy, for example:

```sh
*/assets/characters/*|*/assets/audio/*)
    rel="$${f##*/web/}"   # strips everything up to and including /web/
    dest="$$STAGING/$$rel"
    mkdir -p "$$(dirname "$$dest")" && cp "$$f" "$$dest" ;;
```

Or restructure to copy the whole subtree at once after the loop, mirroring what `serve.sh` does.

cgruber: **`asset_files` glob covers all of `assets/**/*`, but genrule only dispatches on `characters/` and `audio/`**

The `filegroup` uses `glob(["assets/**/*"])`, which will include anything added under `assets/` in the future. The genrule's case statement only handles `*/assets/characters/*` and `*/assets/audio/*`. Files in any other subdirectory (e.g. `assets/icons/`, `assets/fonts/`, or even a stray `assets/README.md`) will be silently omitted from `deployable.zip` while appearing correctly in dev serve and e2e runfiles.

The PR comment "Populated incrementally by GAME-060 / GAME-061" suggests only those two dirs are in scope right now, so this is a forward-looking concern rather than a current bug — but it's an easy silent failure to hit during those follow-on tickets.

Consider either:
- documenting the case-list invariant in a BUILD comment ("add a case here for each new `assets/` subdir"), or
- replacing the per-subdir cases with a single catch-all that preserves relative paths for anything under `assets/` (which also fixes the flattening issue above).

cgruber: Fixed. Replaced the per-file case dispatch for assets with a two-step approach: the loop now collects the assets source root path (`${f%/assets/*}/assets`) from any `*/assets/*` file in SRCS, then after the loop does a single `cp -r "$$ASSETS_SRC/." "$$STAGING/assets/"` — the same pattern used by serve.sh and e2e_test.sh. Verified with `bazel build //web:deployable` + `unzip -l` — the zip now shows `assets/audio/` and `assets/characters/` as subdirectories (tree preserved, not flattened).

cgruber: Added a comment to the `asset_files` filegroup. One note: the prescribed comment text ("add a case for each new subdirectory, or this file is silently omitted") described the old per-subdir case dispatch behavior. Since Issue 1 replaced that with a `cp -r` of the whole assets tree, the warning no longer applies — any new subdir is now auto-included. I updated the comment to reflect the new behavior: "new subdirectories are picked up automatically — no changes to BUILD.bazel required when adding new subdirs."

PR #194, opened 2026-05-02
feat(GAME-064): audio playback infrastructure — preload, mute toggle, autoplay handling

### LINKED ISSUE #193: Audio playback infrastructure — preload, mute toggle, autoplay handling
Audio playback infrastructure — preload, mute toggle, autoplay handling

**Ticket:** [GAME-064](https://github.com/cgruber/redistricting-sim/blob/12cf2b3403620c0f5e9fccad7eee9f7ca1acef5c/thoughts/shared/tickets/GAME-064-audio-playback-infrastructure.md)

cgruber:
## Summary

- Adds `game/web/src/audio/audioPlayer.ts` — the `AudioPlayer` module exposing `preload()`, `play()`, `setMuted()`, and `isMuted()`.
- `play()` silently no-ops when muted, when the clip name is unknown, or when the browser blocks autoplay (catches rejected Promise from `HTMLAudioElement.play()`).
- `setMuted()` persists to localStorage key `redistricting-sim-audio-muted`; `isMuted()` reads it on first access, defaulting to `false`.
- Wires `//web/src/audio:audioplayer_lib` into `app_typecheck_lib` and the esbuild `app` bundle in `game/web/BUILD.bazel`.

- N/A — infrastructure-only; no audio files included, no runtime behavior visible to end users until GAME-062 calls into this module.

## Validation performed

- `bazel test //web/src/audio:all` — 2/2 tests pass (`audioplayer_test`, `audioplayer_typecheck_test`).
- `bazel build //web:app` — esbuild bundle builds cleanly with the new dep wired in.
- `bazel test //web:app_typecheck_test` — type-check passes.
- Node localStorage and Audio shims are set up before module import, matching the pattern in `progress_wip_test.ts`.

## Deployment steps

- N/A — static asset; deployed as part of normal game bundle.

## Tickets addressed

- Closes #193 (GAME-064: Audio playback infrastructure)

## Rollback

- Revert this PR. The module is not called by any existing code so rollback has no user-visible effect.

### DIFF
--- a/game/web/src/audio/audioPlayer_test.ts
+++ b/game/web/src/audio/audioPlayer_test.ts
@@ -1,0 +1,213 @@
+/**
+ * Unit tests for audioPlayer.ts (GAME-064).
+ *
+ * Uses the shared TAP runner. Run via Bazel:
+ *   bazel test //web/src/audio:audioplayer_test
+ *
+ * Neither `localStorage` nor `Audio` exist in Node. Both are shimmed on
+ * globalThis before the module under test is imported so the audio functions
+ * hit the shims rather than missing globals.
+ *
+ * Coverage:
+ *   mute toggle:
+ *     - setMuted(true) → isMuted() === true
+ *     - setMuted(false) → isMuted() === false
+ *     - mute state round-trips through localStorage (persisted + reloaded)
+ *   play():
+ *     - no-ops when muted (no error thrown; Audio.play() never called)
+ *     - no-ops for unknown clip name (no error thrown)
+ *     - silently no-ops when play() returns a rejected Promise (autoplay policy)
+ *     - silently no-ops when play() throws synchronously (older browsers)
+ *   preload():
+ *     - registers clip names; play() of unknown name → silent no-op
+ *     - play() of registered clip attempts playback when unmuted
+ */
+
+// ─── localStorage shim (must be set up before importing audioPlayer.ts) ──────
+
+const _store: Map<string, string> = new Map();
+
+const localStorageShim = {
+  getItem(key: string): string | null {
+    return _store.get(key) ?? null;
+  },
+  setItem(key: string, value: string): void {
+    _store.set(key, value);
+  },
+  removeItem(key: string): void {
+    _store.delete(key);
+  },
+  clear(): void {
+    _store.clear();
+  },
+};
+
+(globalThis as unknown as Record<string, unknown>)["localStorage"] = localStorageShim;
+
+// ─── Audio element shim ───────────────────────────────────────────────────────
+
+/** Tracks whether play() was called on any shim instance. */
+let _audioPlayCalled = false;
+
+/**
+ * Controls what the next AudioShim.play() call returns.
+ * Default: Promise.resolve() (normal success).
+ * Tests that exercise autoplay rejection or synchronous throw set this
+ * to the desired failure mode before calling play().
+ */
+let _nextPlayResult: "resolve" | "reject" | "throw" = "resolve";
+
+class AudioShim {
+  src: string;
+  preload: string = "auto";
+
+  constructor(src: string) {
+    this.src = src;
+  }
+
+  play(): Promise<void> {
+    _audioPlayCalled = true;
+    if (_nextPlayResult === "throw") {
+      throw new DOMException("NotAllowedError", "NotAllowedError");
+    }
+    if (_nextPlayResult === "reject") {
+      return Promise.reject(new DOMException("NotAllowedError", "NotAllowedError"));
+    }
+    return Promise.resolve();
+  }
+}
+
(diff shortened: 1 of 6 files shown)


game/web/src/audio/audioPlayer_test.ts
```diff
@@ -0,0 +1,178 @@
...
+let _audioPlayCalled = false;
+
+class AudioShim {
+  src: string;
+  preload: string = "auto";
+
+  constructor(src: string) {
+    this.src = src;
+  }
+
```
cgruber, line 67: The `AudioShim.play()` here always returns `Promise.resolve()`, so the most important new behavior in this PR — `play()` silently no-op'ing on a rejected Promise / DOMException from the browser's autoplay policy — is never actually exercised by any test. The behavioral contract in `audioPlayer.ts` explicitly lists this as a documented guarantee, and `play()` has two catch sites for it (the inner `result.catch(() => {})` for the modern Promise path and the outer `try { ... } catch {}` for the synchronous-throw path on older browsers). Both are dead from a coverage perspective.

Suggest adding two cases:

1. A shim variant whose `play()` returns `Promise.reject(new Error("NotAllowedError"))` — assert that calling `play("name")` does not throw and that the unhandled-rejection event is not raised. If a future refactor drops the `.catch()`, Node would surface an unhandled rejection and (on `--unhandled-rejections=throw`) fail the test, which is exactly the regression signal needed.
2. A shim variant whose `play()` throws synchronously — assert `play()` does not throw.

A small ergonomic option: parameterize `AudioShim`'s play behavior (a `_nextPlayResult` module variable similar to `_audioPlayCalled`) so individual tests can opt into resolve / reject / throw without redefining the class.

game/web/src/audio/audioPlayer_test.ts
```diff
@@ -0,0 +1,178 @@
...
+import { test, assertEqual, assertTrue, assertFalse, assertDoesNotThrow, summarize } from "../testing/test_runner.js";
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+function resetAll(): void {
+  _store.clear();
+  _audioPlayCalled = false;
+  // Reset the module's internal mute cache by calling setMuted then clearing
+  // storage — we need a way to re-read localStorage. We reset by setting
+  // muted to false (which also writes to storage), then clear storage so
```
cgruber: `resetAll()` clears `_store` and the mute cache, but the module-level `_clips` Map inside `audioPlayer.ts` is never reset. The clip registry accumulates entries across every test in the suite, so test isolation here depends on tests never registering colliding names.

Concretely, the "play() of unknown clip name" tests rely on `"nonexistent"` not being registered. That works today only because nothing else preloads it. If a future test ever calls `preload({ nonexistent: ... })` (or a future contributor renames a fixture), the unknown-name tests would silently start exercising the registered path instead — and still pass, because `Audio.play()` is also a no-op.

Two ways out:

- Export a test-only `_resetForTesting()` from `audioPlayer.ts` that clears `_clips` and resets `_mutedCache` to `undefined`. `resetAll()` here calls it. This also cleans up the awkward "setMuted(false) then clear storage" workaround noted in the `resetAll` comment.
- Or, restructure `audioPlayer.ts` to expose a factory (`createAudioPlayer()`) that returns a fresh instance; production code constructs a singleton, tests construct fresh instances per case.

The `_resetForTesting` approach is the smaller change and is consistent with how the rest of the codebase tends to handle module singletons.

game/web/src/audio/audioPlayer_test.ts
cgruber: Fixed. Added a `_nextPlayResult` control variable (`"resolve" | "reject" | "throw"`, defaulting to `"resolve"`) to `AudioShim`. Added two new tests:

- `"play() silently no-ops when browser rejects play() Promise (autoplay policy)"` — sets `_nextPlayResult = "reject"`, calls `play()`, asserts no synchronous throw. The `.catch()` handler in `audioPlayer.ts` silences the rejection.
- `"play() silently no-ops when play() throws synchronously (older browser)"` — sets `_nextPlayResult = "throw"`, asserts no throw propagates out of `play()`.

`resetAll()` now resets `_nextPlayResult` back to `"resolve"` so the existing tests are unaffected.

game/web/src/audio/audioPlayer_test.ts
cgruber: Fixed. Exported `_resetForTesting()` from `audioPlayer.ts` — it calls `_clips.clear()` and sets `_mutedCache = undefined`. `resetAll()` in the test file now calls `_resetForTesting()` instead of the previous workaround (`setMuted(false)` + `_store.clear()`). This gives each test a genuinely clean clip registry and a null mute cache that re-reads from localStorage on next access.

PR #212, opened 2026-05-07
feat(GAME-062, GAME-067): character reaction system + asset URL versioning

cgruber:
## Summary

**GAME-062 — Governor sprite on result screen**
- `narrative.instigator` field added to scenario format (`InstigatorType` union, validated in loader); scenario-002 declares `"type": "governor"`
- Governor PNG sprite sheet wired to result screen; pose (neutral/approve/disapprove) chosen from star count; element width set per-pose to eliminate cross-column bleed
- Flood-fill background removal with white-matte de-matting: BFS from edges protects interior whites (teeth, highlights); contour edges get correct alpha + de-matted RGB, eliminating halo on dark backgrounds
- `tools/flood-transparent.main.kts`: new reusable BFS-from-edges transparency tool
- `web/asset-preview.html`: local dev-only crop preview (not deployed, not in BUILD)
- DESIGN-010 and GAME-066 filed for the next result screen phase

**GAME-067 — Asset URL versioning + dev/staging version badge**
- `release.main.kts` now patches `app-version` / `app-environment` `<meta>` tags in `index.html` at deploy time (alongside the existing `bundle.js` cache-bust); no network fetch needed in JS, no 404 console errors
- `src/assets.ts`: `initAssets()` reads meta tags synchronously; `assetUrl()` appends `?v=<version>` to all static asset paths; version badge rendered in bottom-left on non-production hostnames
- Eliminates stale-cache hits for PNGs and audio after deploys

N/A — static browser game; no server-side changes

## Validation performed

- [x] TypeScript type check passes (`bazel test //web:app_typecheck_test`)
- [x] 126/126 e2e tests pass (`bazel test //web:e2e_test`)
- [x] e2e: governor sprite visible with `.character-governor` class on result screen after submit
- [x] Verified on dev.pastthepost.gg: all three demographic variants render with transparent backgrounds, correct pose crops, and `?v=` versioned URLs
- [x] Version badge visible bottom-left on dev with correct env + version string
- [x] Teeth and eye highlights remain opaque (BFS flood-fill protects interior whites)
- [x] Sprite bob animation plays; `prefers-reduced-motion: reduce` suppresses it
- [x] No badge rendered on production hostname (by hostname check)
- [x] Non-instigator scenarios: emoji fallback still works
- [x] No console errors when `deployment-metadata.json` absent (e2e test env)

## Deployment steps

POST-MERGE — cut a versioned release and deploy to staging per the normal release workflow

## Tickets addressed

- see GAME-062
- see GAME-067

## Rollback

Revert the PR; the previous commit had no character reaction system (emoji fallback) and no asset versioning


game/release.main.kts
```diff
@@ -364,22 +364,36 @@ class Deploy : CliktCommand(
...
+                html = bundlePatched
+
+                html = html.replace(
+                    """<meta name="app-version" content="">""",
+                    """<meta name="app-version" content="$version">"""
+                )
+                html = html.replace(
+                    """<meta name="app-environment" content="">""",
+                    """<meta name="app-environment" content="$env">"""
+                )
```
cgruber, line 401: The two new meta-tag patches have no effect-check, unlike the bundle.js patch directly above them which warns when its replacement string isn't found. If `index.html` is ever modified so the `content=""` attribute form changes (e.g. single quotes, or the tag is removed), deploys will silently succeed with empty metadata: `assetUrl()` will never append `?v=`, the version badge will show "no metadata" on dev/staging, and there's no operator signal that anything went wrong.

Suggest adding parallel guards, mirroring the bundle.js warning pattern:

```kotlin
val versionPatched = html.replace(
    """<meta name="app-version" content="">""",
    """<meta name="app-version" content="$version">"""
)
if (versionPatched == html) {
    System.err.println("⚠ WARNING: app-version meta tag patch had no effect — tag not found in index.html")
    System.err.println("  Asset cache-busting (?v=) will not work. Check index.html and update release.main.kts.")
}
html = versionPatched

val envPatched = html.replace(
    """<meta name="app-environment" content="">""",
    """<meta name="app-environment" content="$env">"""
)
if (envPatched == html) {
    System.err.println("⚠ WARNING: app-environment meta tag patch had no effect — tag not found in index.html")
}
html = envPatched
```

game/web/src/assets.ts
```diff
@@ -0,0 +1,65 @@
...
+ * On non-production hostnames (dev / staging / localhost) a small version badge
+ * is painted on screen so it is always obvious which build is being tested.
+ */
+
+let _version: string | null = null;
+let _environment: string | null = null;
+
+/**
+ * Read deployment metadata from <meta> tags and show the version badge on
+ * non-production hosts. Synchronous and infallible — safe to await at startup.
```
cgruber: The doc comment says "Synchronous and infallible — safe to await at startup." but `initAssets()` returns `void`, not `Promise<void>`. `await`-ing a void is harmless, but the wording reads like a leftover from an earlier async design (the GAME-067 ticket still describes fetching `deployment-metadata.json` via network). Since the implementation is now sync and reads from `<meta>` tags, the comment should say "Synchronous — safe to call at the top of the async IIFE" (or simply drop the "safe to await" phrasing).

game/web/src/assets.ts
```diff
@@ -0,0 +1,65 @@
...
+ * fresh content after a deploy.
+ */
+export function assetUrl(path: string): string {
+  return _version !== null ? `${path}?v=${_version}` : path;
+}
+
+// ─── Internal helpers ─────────────────────────────────────────────────────────
+
+function isProduction(): boolean {
+  return window.location.hostname === "pastthepost.gg";
```
cgruber: nit: `isProduction()` matches only `pastthepost.gg` exactly. If `www.pastthepost.gg` is ever a live hostname (common for apex-vs-www setups), the version badge will show on production. Not a bug today given the current DNS setup, but worth a comment acknowledging this, or broadening the check:

```typescript
function isProduction(): boolean {
  const h = window.location.hostname;
  return h === "pastthepost.gg" || h === "www.pastthepost.gg";
}
```

game/release.main.kts
cgruber: Fixed. Added no-effect warnings for both meta-tag patches, matching the style of the bundle.js warning:

```kotlin
val versionMetaOriginal = html
html = html.replace(...)
if (html == versionMetaOriginal) {
    System.err.println("⚠ WARNING: app-version meta patch had no effect — ...")
    System.err.println("  The JS version badge and asset URLs may show stale or missing data. ...")
}

val envMetaOriginal = html
html = html.replace(...)
if (html == envMetaOriginal) {
    System.err.println("⚠ WARNING: app-environment meta patch had no effect — ...")
    System.err.println("  The JS version badge may show the wrong environment. ...")
}
```

Each patch snapshots `html` before the replace and compares after, so a no-op is caught regardless of what earlier patches did.

game/web/src/assets.ts
cgruber: Fixed. Changed "safe to await at startup" → "call once at startup". The function returns `void` so "await" was misleading — callers don't await it. The rest of the description (synchronous, infallible) still holds.

game/web/src/assets.ts
cgruber: Fixed. Updated `isProduction()` to match both bare and `www.` hostname:

```ts
function isProduction(): boolean {
  const h = window.location.hostname;
  return h === "pastthepost.gg" || h === "www.pastthepost.gg";
}
```

This ensures the version badge is suppressed on the www subdomain the same as the apex domain, so it won't appear in production if traffic is ever routed through `www.`.

PR #214, opened 2026-05-07
fix(GAME-068): true sequential criterion reveal + dedicated Skip button

cgruber:
## Prerequisites
- [x] N/A — no parent PR

## Summary

Fixes the reveal pacing from GAME-066: the 400ms stagger caused rows to pile up in
CHECKING state simultaneously (wave of criteria → wave of results). This changes to
a true chain where each row's full cycle completes before the next starts.

- **Chain delay**: 300ms fade + 1200ms CHECKING hold + 150ms flip = 1650ms per row;
  next row starts only after current row has fully resolved — no simultaneous CHECKING
- **Skip button**: dedicated `#btn-reveal-skip` button replaces click-anywhere-to-skip;
  visible only during animated reveal, hidden in reduced-motion path and after completion
- Sprint4 GAME-052 test updated to reflect that skip is now via button, not screen click

## Validation performed

- [x] `bazel build //web:app_typecheck_lib` passes
- [x] `bazel test //web:e2e_test` — 129/129 pass
- [x] GAME-068 skip button test: visible during animated reveal, hides after skip
- [x] GAME-052 test updated: reduced-motion instant-reveal + no skip button shown

- Browser game only (`game/web/`) — client-side JS/CSS, no server-side changes

- Deploy to dev: `./release.main.kts -- deploy --env dev --version <version>` from `game/`

## Tickets addressed

- see GAME-068

## Rollback

- Revert commit on main; redeploy — no data migrations, fully reversible

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -95,6 +95,8 @@ const resultReaction = document.getElementById("result-reaction") as HTMLElement
 const resultCriteriaList = document.getElementById("result-criteria-list") as HTMLElement | null;
 const btnKeepDrawing = document.getElementById("btn-keep-drawing") as HTMLButtonElement | null;
 const btnNextScenario = document.getElementById("btn-next-scenario") as HTMLButtonElement | null;
+const resultRevealControls = document.getElementById("result-reveal-controls") as HTMLElement | null;
+const btnRevealSkip = document.getElementById("btn-reveal-skip") as HTMLButtonElement | null;
 
 // Intro screen refs (GAME-016)
 const introScreen = document.getElementById("intro-screen") as HTMLElement | null;
@@ -806,7 +808,7 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 
 	function showResultScreen() {
 		if (!resultScreen || !resultVerdict || !resultSubtitle || !resultCriteriaList || !resultReaction) return;
-		if (skipClickHandler) { resultScreen.removeEventListener("click", skipClickHandler); skipClickHandler = null; }
+		if (skipClickHandler) { btnRevealSkip?.removeEventListener("click", skipClickHandler); skipClickHandler = null; }
 
 		const state = store.getState();
 		if (state.simulationResult === null) return;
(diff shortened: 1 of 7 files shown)


game/web/src/main.ts
```diff
@@ -976,8 +990,8 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 						// After last row resolves, pause then show instigator verdict.
 						if (i === rowElements.length - 1) {
 							const tVerdict = setTimeout(() => {
-								// Natural reveal complete — deactivate skip handler.
-								resultScreen.removeEventListener("click", skipHandler);
+								// Natural reveal complete — deactivate skip.
+								if (resultRevealControls) resultRevealControls.style.display = "none";
 								skipClickHandler = null;
```
cgruber, line 996: **Bug: stale `btnRevealSkip` listener accumulates across result-screen re-opens**

Two gaps together cause a double-fire on the second Skip click:

**Gap 1 — natural completion doesn't remove the listener (line ~994-995).**

When the reveal finishes naturally, the code hides the controls and sets `skipClickHandler = null`:

```ts
if (resultRevealControls) resultRevealControls.style.display = "none";
skipClickHandler = null;
```

But the `{once:true}` listener on `btnRevealSkip` was never fired — it is still attached. `{once:true}` only self-removes on the first invocation; no-click = still live.

**Gap 2 — the cleanup guard on re-open (line 811) targets the wrong element.**

```ts
if (skipClickHandler) { resultScreen.removeEventListener("click", skipClickHandler); skipClickHandler = null; }
```

The listener now lives on `btnRevealSkip`, not `resultScreen`. This `removeEventListener` call is a no-op. Additionally, if natural completion ran first, `skipClickHandler` is already null, so the guard doesn't fire at all.

**Failure path:**

1. Submit → animated reveal plays to completion naturally (no Skip click).
2. User clicks Keep Drawing → result screen hidden.
3. User resubmits → `showResultScreen()` called again.
4. Line 811 guard: `skipClickHandler` is null → skips cleanup (and would target wrong element anyway).
5. Line 1026: new listener added. `btnRevealSkip` now has two `{once:true}` listeners.
6. User clicks Skip → both fire: old closure calls `transitionInstigatorToVerdict` with stale `waitingSprite`/`rowElements`; new closure does it correctly. Double verdict transition.

**Fix — two changes needed:**

At natural completion, remove the listener explicitly:

```ts
// Natural reveal complete — deactivate skip.
if (resultRevealControls) resultRevealControls.style.display = "none";
btnRevealSkip?.removeEventListener("click", skipHandler);  // ← add this
skipClickHandler = null;
```

And update the re-open cleanup at line 811 to target the right element:

```ts
if (skipClickHandler) {
  btnRevealSkip?.removeEventListener("click", skipClickHandler);  // ← was resultScreen
  skipClickHandler = null;
}
```

cgruber: Fixed both gaps:
1. `tVerdict` now calls `btnRevealSkip?.removeEventListener("click", skipHandler)` before hiding controls — stale listener cleared on natural completion.
2. Re-open guard (line ~811) changed from `resultScreen.removeEventListener(...)` to `btnRevealSkip?.removeEventListener(...)` — targets the correct element.

PR #218, opened 2026-05-13
feat: add image-tools scripts, merge flood-fill modes, wire CI test step

cgruber:
## Prerequisites
- [x] N/A — no parent PR

## Summary
- Merges `flood-fill-alpha.main.kts` (seed-point mode) and `flood-transparent.main.kts` (edge mode) into a single `flood-fill.main.kts` with `--seed x,y` (repeatable) and `--edge` flags; flags are combinable; default is edge mode (preserving prior behavior with default tolerance 75)
- Adds `find-white.main.kts` (merged from two `/tmp` analysis scripts): count + sample coords by default, `--spatial` flag for 50px grid bucketing
- Adds `check-pixels.main.kts`: inspect ARGB values at specific x,y coordinates
- Adds `sample-colors.main.kts`: find pixels matching a configurable color filter (default: pinkish/magenta), parameterized via `--min-r`/`--max-g`/`--min-b`
- Adds `show-seed-points.main.kts`: overlay annotated crosshair markers on a PNG for visual debugging of seed coordinates
- Adds `normalize-magenta.main.kts`: normalize magenta/pink color values in sprites
- Adds `tools/image-tools.test.main.kts`: JUnit Jupiter unit tests covering pure logic from all new scripts (`FloodLogic`, `FindWhiteLogic`, `SampleColorsLogic`)
- Adds `:kotlin: Tools Tests` CI step to `.buildkite/pipeline.yml` — globs `tools/*.test.main.kts` (picks up existing `gen-assets.test.main.kts` too), runs independently of the Bazel build

## Validation performed
- [ ] N/A — no kubectl/ansible changes
- [ ] N/A — no SOPS secrets
- Logic extracted into `object` blocks in each script so `image-tools.test.main.kts` can test it without a build system
- Test file mirrors the pattern established by `gen-assets.test.main.kts`

- Buildkite CI (macOS agent): new Tools Tests step runs `kotlin` on `tools/*.test.main.kts`; `kotlin` confirmed available at `/opt/homebrew/bin/kotlin` on the agent

- Automatic — scripts are standalone `.main.kts` files on PATH; CI step activates on next build

## Tickets addressed
- N/A

## Rollback
- Revert this PR; the deleted `flood-fill-alpha.main.kts` and `flood-transparent.main.kts` are recoverable from git history


.buildkite/pipeline.yml
```diff
@@ -51,3 +51,18 @@ steps:
...
+  - label: ":kotlin: Tools Tests"
+    key: tools-test
+    agents:
+      os: macos
+    command: |
+      set -euo pipefail
+      for f in tools/*.test.main.kts; do
+        echo "--- $f"
+        kotlin "$f"
+      done
```
cgruber, line 70: The new `tools-test` step calls `kotlin "$f"` directly, but `kotlin` (the Kotlin script runner / `kotlinc`) is not installed by the existing pipeline steps and doesn't appear in any agent setup block. The existing `build` and `test` steps only invoke `bazel`, which manages its own JDK toolchain internally.

If the macOS agent doesn't have `kotlin` on `PATH` this step will fail with `command not found` on first run. Please either:
- Document the agent precondition (e.g. `sdk install kotlin` / `brew install kotlin` in agent bootstrap), or
- Add an install step, or
- Confirm in a comment that `kotlin` is already present on the agent image.

tools/normalize-magenta.main.kts
```diff
@@ -0,0 +1,50 @@
...
+            out.setRGB(x, y, (0xFF shl 24) or 0xFF00FF)
+            count++
+        } else {
+            out.setRGB(x, y, (0xFF shl 24) or (px and 0xFFFFFF))
+        }
+    }
+}
+
+ImageIO.write(out, "PNG", File(output))
+println("Normalized $count pixels → $output")
```
cgruber, line 57: Both branches of the pixel loop force `alpha = 0xFF`, which silently destroys any existing transparency in the source image:

```kotlin
out.setRGB(x, y, (0xFF shl 24) or 0xFF00FF)  // hot-pink → pure magenta, alpha forced opaque
out.setRGB(x, y, (0xFF shl 24) or (px and 0xFFFFFF))  // all other pixels, alpha also forced opaque
```

The file header says "Run this before transparency fill", but that precondition isn't enforced or mentioned here at the code site. If a caller runs this on an already-processed (partially transparent) PNG — e.g. after a first pass of `flood-fill.main.kts` — all transparency will be silently wiped.

Suggest either:
1. Add an `require` or warning if the source image has any transparent pixels (alpha < 255), or
2. Preserve source alpha in the non-matched branch: `(px and 0xFF000000.toInt()) or (px and 0xFFFFFF)`, or
3. Add a comment directly above the `setRGB` calls documenting the "opaque-only input" precondition explicitly.

tools/show-cutlines.main.kts
```diff
@@ -0,0 +1,41 @@
...
+val outDir = File(args.getOrElse(1) { "/tmp/cutline-guides" })
+outDir.mkdirs()
+
+val sheets = assetsDir.listFiles()!!
+    .filter { it.isDirectory }
+    .mapNotNull { dir -> File(dir, "sheet.png").takeIf { it.exists() }?.let { dir.name to it } }
+    .sortedBy { it.first }
+
+for ((name, file) in sheets) {
+    val img = ImageIO.read(file)
```
cgruber, line 18: `listFiles()` returns `null` (not an empty list) when `assetsDir` either doesn't exist or isn't a directory. The `!!` here gives a `NullPointerException` with no context about which path caused the problem.

Suggest replacing with a `requireNotNull` or an explicit check:

```kotlin
val files = assetsDir.listFiles()
    ?: error("Cannot list files in: ${assetsDir.absolutePath} (does not exist or is not a directory)")
```

This makes the failure message actionable instead of just a stack trace.

tools/normalize-magenta.main.kts
```diff
@@ -0,0 +1,50 @@
...
+
+val input  = args.getOrNull(0) ?: error("Usage: normalize-magenta.main.kts <input.png> <output.png> [tolerance]")
+val output = args.getOrNull(1) ?: error("Usage: normalize-magenta.main.kts <input.png> <output.png> [tolerance]")
+val tol    = args.getOrNull(2)?.toInt() ?: 80
+
+val src = ImageIO.read(File(input))
+val out = BufferedImage(src.width, src.height, BufferedImage.TYPE_INT_ARGB)
+
+// Detected hot-pink center from sampling: R≈250, G≈0, B≈130
+// Criteria: high R, low G, mid-range B (distinguishes from true blue or true magenta)
```
cgruber, line 27: `args.getOrNull(2)?.toInt()` will throw an unguarded `NumberFormatException` if the caller passes a non-integer tolerance (e.g. `./normalize-magenta.main.kts in.png out.png abc`). The error message will be something like `For input string: "abc"` with no reference to the argument name.

Suggest:
```kotlin
val tol = args.getOrNull(2)?.toIntOrNull()
    ?: args.getOrNull(2)?.let { error("Tolerance must be an integer, got: $it") }
    ?: 80
```

Or more concisely:
```kotlin
val tol = args.getOrNull(2)?.let {
    it.toIntOrNull() ?: error("Tolerance must be an integer, got: $it")
} ?: 80
```

.buildkite/pipeline.yml
cgruber: Fixed. Added a comment block above the `for` loop documenting that `kotlin` is a known agent precondition, confirmed present at `/opt/homebrew/bin/kotlin` on the macOS dev-machine agent, and that no install step is needed since this is a persistent agent rather than ephemeral CI.

tools/normalize-magenta.main.kts
cgruber: Fixed with option (a). The pixel loop now extracts the alpha channel first and skips fully transparent pixels (`a == 0`) by copying them as-is. For opaque/semi-transparent non-hot-pink pixels, the original alpha is now preserved (`(a shl 24) or (px and 0xFFFFFF)`) instead of forcing `0xFF`.

tools/show-cutlines.main.kts
cgruber: Fixed. Replaced `listFiles()!!` with `(assetsDir.listFiles() ?: error("assets directory not found or not a directory: ${assetsDir.absolutePath}"))` so a missing or non-directory path produces a clear error message instead of an NPE.

tools/normalize-magenta.main.kts
cgruber: Fixed. Changed to `args.getOrNull(2)?.let { it.toIntOrNull() ?: error("tolerance must be an integer, got: $it") } ?: 80` so a non-integer argument produces a clear error message instead of an unhandled `NumberFormatException`.

PR #223, opened 2026-05-13
feat(GAME-062): wire commissioner/judge/legislator/party sprites; add character_demographics to scenario

cgruber:
## Prerequisites
- [ ] N/A | Parent PR merged: none

## Summary
- Wires all 5 character types (commissioner, judge, legislator, party, governor) to their PNG sprite sheets on the result screen — previously only governor was wired; the other 4 fell through to placeholder SVGs.
- Adds `character_demographics` optional field to `Scenario` (TypeScript model + JSON loader) so each scenario JSON can specify which demographic variant to show per character type, instead of a random or hash-derived selection.
- All 10 scenario JSON files updated with `character_demographics` mappings (game content files, not infrastructure; classified as D by diff-summary heuristic but contain no new services/pipelines).
- Sprite sheet layout constants (`CHAR_SHEET`, `CHAR_ROW_SCALE`) added for the 1408×768 non-governor sheet format.

## Validation performed
- [ ] N/A | kubectl dry-run passed (N/A — not a k8s change)
- [ ] N/A | sops decrypt verified (N/A — no secrets)
- [x] Bazel build `//web/...` passes cleanly

- Game web app (client-side only; no server changes)
- Scenario JSON files are game content shipped with the web app bundle; no new services or pipelines introduced

- Standard game release via `game/release.main.kts -- deploy --env dev` (branch test) or full semver release from main

## Tickets addressed
- see #TBD (GAME-062)

## Rollback
- Revert this commit; prior behavior showed placeholder SVGs for non-governor characters on the result screen

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -739,13 +739,16 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		return criterionResults.filter(cr => cr.required && cr.passed).length;
 	}
 
-	const GOVERNOR_DEMOGRAPHICS = ["wm", "bm", "af"] as const;
-
-	// GAME-069: Governor sprite sheet layout (200px tall canonical):
-	// Sheet: 1376×752px. Pose columns: neutral 0–400, approve 400–880, disapprove 880–1376.
-	// Row-scale (84px target height): 84/752 ≈ 0.1117.
+	// GAME-069/GAME-062: Sprite sheet layout constants.
+	// Governor sheet: 1376×752px (unique dimensions).
+	// Pose columns: neutral 0–400, approve 400–880, disapprove 880–1376.
 	const GOV_ROW_SCALE = 84 / 752;
 	const GOV_SHEET = { neutral: { x: 0, w: 400 }, approve: { x: 400, w: 480 }, disapprove: { x: 880, w: 496 } };
+
+	// All other character sheets: 1408×768px, equal-thirds columns (469/469/470).
+	const CHAR_ROW_SCALE = 84 / 768;
+	const CHAR_SHEET_WIDTH = 1408;
+	const CHAR_SHEET = { neutral: { x: 0, w: 469 }, approve: { x: 469, w: 469 }, disapprove: { x: 938, w: 470 } };
 
 	// Placeholder SVG for non-governor character types.
 	function charPlaceholderSvg(state: "neutral" | "approve" | "disapprove"): string {
@@ -773,7 +776,6 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		slot: HTMLElement,
 		charType: CharacterType,
 		passed: boolean,
-		demo: string,
 	): { neutral: HTMLElement; verdict: HTMLElement } {
 		const neutralEl = document.createElement("div");
 		neutralEl.className = "rc-char-neutral";
@@ -781,10 +783,14 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		verdictEl.className = "rc-char-verdict";
 		verdictEl.style.opacity = "0";
 
+		// Demographic variant: read from scenario, fall back to "" (generic/no-suffix).
+		const demo = scenario.character_demographics?.[charType] ?? "";
+
 		if (charType === "governor") {
 			const n = GOV_SHEET.neutral;
 			const v = passed ? GOV_SHEET.approve : GOV_SHEET.disapprove;
-			const img = assetUrl(`assets/characters/governor-${demo}/sheet.png`);
+			const govDemo = demo || "wm"; // loader enforces non-empty; fallback guards stale/direct scenarios
+			const img = assetUrl(`assets/characters/governor-${govDemo}/sheet.png`);
 			const makeSprite = (col: { x: number; w: number }, label: string): HTMLElement => {
 				const s = document.createElement("div");
 				s.className = "character-sprite character-governor character-sprite--row";
@@ -799,8 +805,26 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 			neutralEl.appendChild(makeSprite(n, "Character awaiting verdict"));
 			verdictEl.appendChild(makeSprite(v, passed ? "Approves" : "Disapproves"));
 		} else {
-			neutralEl.innerHTML = charPlaceholderSvg("neutral");
-			verdictEl.innerHTML = charPlaceholderSvg(passed ? "approve" : "disapprove");
+			// charType is commissioner | judge | legislator | party — all use 1408×768 sheets.
+			// party has no demographic variants (path: party/sheet.png).
+			// judge allows empty demo → bare judge/ directory; all others require non-empty (enforced by loader).
+			const dir = (charType !== "party" && demo) ? `${charType}-${demo}` : charType;
+			const img = assetUrl(`assets/characters/${dir}/sheet.png`);
+			const n = CHAR_SHEET.neutral;
+			const v = passed ? CHAR_SHEET.approve : CHAR_SHEET.disapprove;
+			const makeSprite = (col: { x: number; w: number }, label: string): HTMLElement => {
+				const s = document.createElement("div");
+				s.className = `character-sprite character-${charType} character-sprite--row`;
+				s.setAttribute("role", "img");
+				s.setAttribute("aria-label", label);
+				s.style.width = `${Math.round(col.w * CHAR_ROW_SCALE)}px`;
+				s.style.backgroundImage = `url('${img}')`;
+				s.style.backgroundPosition = `-${Math.round(col.x * CHAR_ROW_SCALE)}px 0%`;
+				s.style.backgroundSize = `${Math.round(CHAR_SHEET_WIDTH * CHAR_ROW_SCALE)}px 84px`;
+				return s;
+			};
+			neutralEl.appendChild(makeSprite(n, "Character awaiting verdict"));
+			verdictEl.appendChild(makeSprite(v, passed ? "Approves" : "Disapproves"));
 		}
 
 		slot.appendChild(neutralEl);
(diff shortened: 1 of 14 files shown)


game/web/src/main.ts
```diff
@@ -773,14 +776,16 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
-		demo: string,
 	): { neutral: HTMLElement; verdict: HTMLElement } {
 		const neutralEl = document.createElement("div");
 		neutralEl.className = "rc-char-neutral";
 		const verdictEl = document.createElement("div");
 		verdictEl.className = "rc-char-verdict";
 		verdictEl.style.opacity = "0";
 
+		// Demographic variant: read from scenario, fall back to "" (generic/no-suffix).
+		const demo = scenario.character_demographics?.[charType] ?? "";
```
cgruber, line 787: **Governor path will break when `character_demographics.governor` is absent or empty.**

`governor-${demo}` is evaluated unconditionally. When `demo` is `""` (field missing from `character_demographics`, or explicitly set to `""`), the asset URL becomes `assets/characters/governor-/sheet.png`, which produces a 404 and a blank sprite.

The old code always resolved a valid demo from `GOVERNOR_DEMOGRAPHICS` deterministically. The new code can produce an invalid path.

Currently scenario-002 is the only scenario with governor criteria, and it does include `"governor": "bm"` in `character_demographics`, so this isn't triggered yet. But any future scenario that uses `character: "instigator"` (or `character: "governor"`) and omits the `governor` key from `character_demographics` will silently break.

Suggested fix: add a fallback, e.g. `governor-${demo || "wm"}`, or throw/warn at load time if `character_demographics.governor` is absent when any criterion uses the governor character.

game/web/src/main.ts
```diff
@@ -798,6 +803,26 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 			};
 			neutralEl.appendChild(makeSprite(n, "Character awaiting verdict"));
 			verdictEl.appendChild(makeSprite(v, passed ? "Approves" : "Disapproves"));
+		} else if (charType === "commissioner" || charType === "judge" || charType === "legislator" || charType === "party") {
+			// 1408×768 sheets. party has no demographic variants (path: party/sheet.png).
+			// Others: demo suffix if non-empty (e.g. commissioner-wf), else bare dir (e.g. judge).
+			const dir = (charType !== "party" && demo) ? `${charType}-${demo}` : charType;
```
cgruber, line 811: **`legislator` with empty demo will produce a broken path — no bare `legislator/` directory exists.**

The dir-resolution logic correctly falls back to `charType` (bare directory) when `demo` is empty: `(charType !== "party" && demo) ? ... : charType`. For `judge` this is fine because `assets/characters/judge/` exists.

However, only `legislator-bm/`, `legislator-wf/`, and `legislator-wm/` exist in `assets/characters/` — there is no bare `legislator/` directory. If a future scenario uses a criterion with `character: "legislator"` but omits `legislator` from `character_demographics`, demo will be `""`, dir will be `"legislator"`, and the asset URL will 404.

No current scenario triggers this (no criterion currently uses the legislator type), but the invariant is not enforced. Consider either:
1. Creating a bare `legislator/` directory as a default fallback, or
2. Requiring `character_demographics.legislator` at load time when any criterion uses it, similar to how `governor` should be handled.

game/web/src/model/loader.ts
```diff
@@ -587,6 +587,22 @@ export function loadScenario(json: unknown): Scenario {
     instigator_character = icVal as CharacterType;
   }
 
+  let character_demographics: Partial<Record<CharacterType, string>> | undefined;
+  if (raw["character_demographics"] !== undefined) {
+    const cd = raw["character_demographics"];
+    if (typeof cd !== "object" || cd === null || Array.isArray(cd)) {
+      throw new Error("character_demographics: must be an object");
+    }
+    const validChars: string[] = ["governor", "commissioner", "party", "judge", "legislator"];
```
cgruber: **Loader silently accepts `"party"` key in `character_demographics`, but the render path ignores it.**

`validChars` includes `"party"`, so `{"party": "foo"}` is a valid payload from the loader's perspective. But the render code does `charType !== "party"` which means the demo value is ignored for party characters — the path is always `assets/characters/party/sheet.png`.

The JSDoc comment on `Scenario.character_demographics` already says `"party" has no variants and should be omitted`. The loader should enforce this to prevent silently misleading data:

```ts
if (key === "party") {
  throw new Error('character_demographics: "party" has no demographic variants and must be omitted');
}
```

game/web/src/main.ts
cgruber: Fixed. Two-layer fix: (1) loader.ts now validates that governor (and commissioner/legislator) demographic values must be non-empty strings — bare directories don't exist for these types, so an empty value is rejected at parse time with a clear error message. (2) main.ts adds a `govDemo = demo || "wm"` runtime fallback as belt-and-suspenders for any scenario that bypasses the loader (e.g. direct construction in tests).

game/web/src/main.ts
cgruber: Fixed. The loader now enforces non-empty demographic values for `"legislator"` (along with `"governor"` and `"commissioner"`) since none of these have bare directories. Any scenario that omits the key simply won't have a legislator character slot — the empty-demo path is only reachable if a future scenario incorrectly sets `"legislator": ""`, which will now throw at load time.

game/web/src/model/loader.ts
cgruber: Fixed. Removed `"party"` from `validChars` and added an explicit early check that throws a clear error: `"party" has no demographic variants and must be omitted`. This enforces the contract at load time rather than silently discarding the value.

game/web/src/main.ts
```diff
@@ -798,6 +804,26 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
```
cgruber, line 811: Latent 404: asymmetric fallback vs. governor path.

The governor branch above does `demo || "wm"`, so a scenario without a `character_demographics.governor` entry (or with `""`) safely falls back to a real asset directory. This branch has no equivalent guard:

```ts
const dir = (charType !== "party" && demo) ? `${charType}-${demo}` : charType;
```

When `demo` is `""` (key absent → `?? ""` on line above), `dir` becomes `"commissioner"` or `"legislator"` — neither of which has a bare asset directory (`assets/characters/commissioner/sheet.png` does not exist). The loader's `requiresSuffix` check catches this if the key is *present with an empty string*, but the key can simply be *absent* — every scenario with validity rows implicitly displays a commissioner character via the `?? { type: "commissioner" }` fallback.

All 11 scenarios in this PR do supply a non-empty `commissioner` entry, so this is latent rather than currently broken. But it's an easy foot-gun for the next scenario added. Consider mirroring the governor pattern:

```ts
const safeDemo = (charType !== "party")
  ? (demo || defaultDemoFor(charType))  // e.g. "wf" for commissioner, "wf" for legislator
  : "";
const dir = safeDemo ? `${charType}-${safeDemo}` : charType;
```

Or at minimum add a `// NOTE: commissioner and legislator have no bare fallback dir; scenario must supply demographics` comment so future authors know the requirement.

game/web/src/main.ts
```diff
@@ -798,6 +804,26 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
+				s.style.width = `${Math.round(col.w * CHAR_ROW_SCALE)}px`;
+				s.style.backgroundImage = `url('${img}')`;
+				s.style.backgroundPosition = `-${Math.round(col.x * CHAR_ROW_SCALE)}px 0%`;
+				s.style.backgroundSize = `${Math.round(CHAR_SHEET_WIDTH * CHAR_ROW_SCALE)}px 84px`;
+				return s;
+			};
+			neutralEl.appendChild(makeSprite(n, "Character awaiting verdict"));
+			verdictEl.appendChild(makeSprite(v, passed ? "Approves" : "Disapproves"));
 		} else {
 			neutralEl.innerHTML = charPlaceholderSvg("neutral");
```
cgruber: Dead code: this `else` branch is now unreachable.

The preceding `if` covers `"governor"` and the `else if` covers all four remaining `CharacterType` values (`"commissioner" | "judge" | "legislator" | "party"`). `CharacterType` is a closed union — there is no fifth variant that could reach this branch.

The `charPlaceholderSvg` function and its constants are therefore also dead. Not a bug, but they add noise and can be removed (or kept intentionally as a "never" safety net with a comment).

game/web/src/model/scenario.ts
```diff
@@ -256,6 +256,10 @@ export interface Scenario {
 	narrative: Narrative;
 	/** Which character type plays the instigator role in this scenario. Resolves "instigator" character refs on criteria. */
 	instigator_character?: CharacterType;
+	/** Demographic variant per character type for result-screen sprites.
+	 *  Key = CharacterType; value = demographic suffix (e.g. "wm", "bm", "af", "lm", "naf", "wf", "bf", or "" for the
```
cgruber: JSDoc says `""` is "the generic/default variant" but the loader rejects empty strings for governor, commissioner, and legislator.

```
Key = CharacterType; value = demographic suffix (e.g. "wm", "bm", "af", ..., or "" for the
generic/default variant)
```

`""` is only valid for `judge` (the only type with a bare `judge/` asset directory). For `governor`, `commissioner`, and `legislator` the loader throws `demographic suffix must be non-empty`. The comment should clarify which types accept `""`:

```ts
/** Demographic variant per character type for result-screen sprites.
 *  Key = CharacterType (excluding "party"); value = demographic suffix (e.g. "wm", "bf", "lm",
 *  "naf"…). Use "" only for types that have a bare asset directory (currently: "judge").
 *  Governor, commissioner, and legislator require a non-empty suffix.
 *  "party" has no variants and must be omitted entirely. */
```

game/web/src/main.ts
cgruber: Fixed. Collapsed the `else if` for non-governor types into a plain `else` (since governor is the only other branch and all remaining CharacterType values are in that arm). Updated the comment to explicitly document that judge is the only type permitting an empty suffix, and that loader enforces non-empty for the others.

game/web/src/main.ts
cgruber: Fixed. The `else if (charType === "commissioner" || ...)` is now a plain `else`, eliminating the dead branch entirely. TypeScript will still exhaustiveness-check the outer `if (charType === "governor")` vs `else` correctly since CharacterType is a closed union.

game/web/src/model/scenario.ts
cgruber: Fixed. JSDoc now says: "Empty string is valid only for 'judge' (resolves to the bare judge/ directory); governor/commissioner/legislator require a non-empty suffix."

PR #232, opened 2026-05-18
fix(UX): remove About button from scenario-select screen

cgruber:
## Prerequisites
- [x] Parent PR merged: #231

## Summary
- Removes the duplicate About button from the scenario-select screen
- About remains accessible from the main menu (`btn-main-about`), which is the appropriate entry point
- Removes the now-dead `btn-about` / `btn-about-close` event listeners from `showScenarioSelect()`

## Validation performed
- [ ] N/A — kubectl dry-run (not infra)
- [ ] N/A — sops decrypt (not infra)
- Verified: main menu About button and close flow unaffected; scenario-select no longer shows About

- Web game only

## Tickets addressed
- N/A (minor UX cleanup)

## Rollback
- Revert this commit; the button and listeners are self-contained

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -368,15 +368,6 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 			renderScenarioCards();
 		});
 
-		// About button
-		document.getElementById("btn-about")?.addEventListener("click", () => {
-			scenarioSelectEl?.classList.add("hidden");
-			document.getElementById("about-screen")?.classList.remove("hidden");
-		});
-		document.getElementById("btn-about-close")?.addEventListener("click", () => {
-			document.getElementById("about-screen")?.classList.add("hidden");
-			scenarioSelectEl?.classList.remove("hidden");
-		});
 	}
 
 	// ── Startup routing (GAME-021 / GAME-048) ────────────────────────────────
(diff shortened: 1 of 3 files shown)


game/web/src/main.ts
```diff
@@ -368,15 +368,6 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
-		// About button
-		document.getElementById("btn-about")?.addEventListener("click", () => {
-			scenarioSelectEl?.classList.add("hidden");
-			document.getElementById("about-screen")?.classList.remove("hidden");
-		});
-		document.getElementById("btn-about-close")?.addEventListener("click", () => {
-			document.getElementById("about-screen")?.classList.add("hidden");
-			scenarioSelectEl?.classList.remove("hidden");
-		});
 	}
```
cgruber, line 371: The listener removal is correct, but the e2e test that covers this flow was not updated.

`game/web/e2e/scenarios.spec.ts` line 884 — `"about page: accessible from select screen and shows educational content"` — navigates to the scenario-select screen and clicks `#btn-about` (line 891), which no longer exists. It also clicks `#btn-about-close` (line 896) and asserts that `#scenario-select` becomes visible again — but `#btn-about-close` now navigates back to the **main menu**, so that assertion would also fail.

This test needs to be rewritten to reflect the new flow: navigate to `/`, click `#btn-main-about`, verify content, click `#btn-about-close`, verify main menu is visible (not scenario-select). The test title should also be updated since About is no longer accessible from the select screen.

Without this fix, CI will fail.

cgruber: Fixed. Updated the e2e test to navigate via the main menu About button instead: loads `/`, waits for `#main-menu`, clicks `#btn-main-about`, verifies content, clicks `#btn-about-close`, verifies main menu visible. Squashed into the same commit and pushed.

PR #233, opened 2026-05-18
feat(debug): force-win plays result screen; per-criterion pass/fail replay

cgruber:
## Prerequisites
- [x] Parent PR merged: #232

## Summary
- Debug force-win (`⚡ Force Win`) now opens the result screen with all criteria forced to pass, running the full sequential reveal animation instead of navigating away — lets animators and testers verify the reveal flow without solving a map
- In debug mode (`?debug`), each criterion row gains `✓ pass` and `✗ fail` buttons; clicking either cancels any in-flight animation and re-runs the single-row reveal with the chosen verdict (audio + character sprite transition included), enabling per-criterion audio and animation testing in isolation
- Shared `activeTimeouts` array replaces the previous `pendingTimeouts` local so both the main reveal, skip handler, debug replay, and Keep Drawing all cancel from the same source — no more orphaned audio after clicking Fix It during a replay

## Validation performed
- [ ] N/A — kubectl dry-run (not infra)
- [ ] N/A — sops decrypt (not infra)
- TypeScript typecheck passes (`bazel test //game/web:app_typecheck_test`)
- Built clean (`bazel build //game/web:app`)
- Manually verified: Force Win opens result screen with all-green criteria reveal; pass/fail buttons replay individual rows with correct audio

- Web game only (debug-mode behaviour; no change to production path)

## Tickets addressed
- N/A (debug tooling improvement)

## Rollback
- Revert this commit; changes are self-contained to `showResultScreen` and the debug win button

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -880,7 +880,7 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		verdict.style.opacity = "1";
 	}
 
-	function showResultScreen() {
+	function showResultScreen(debugForcePass = false) {
 		if (!resultScreen || !resultVerdict || !resultSubtitle || !resultCriteriaList) return;
 		if (skipClickHandler) { btnRevealSkip?.removeEventListener("click", skipClickHandler); skipClickHandler = null; }
 
@@ -895,7 +895,7 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		);
 
 		// GAME-059: detect invalid map before running criterion evaluation.
-		const mapIsValid = isMapSubmittable(validity, scenario.rules);
+		let mapIsValid = isMapSubmittable(validity, scenario.rules);
 
 		const evalResult = evaluateCriteria(
 			scenario.success_criteria,
@@ -910,7 +910,12 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		);
 
 		// GAME-059: overall pass requires a valid map AND all required criteria passing.
-		const overallPass = mapIsValid && evalResult.overallPass;
+		let overallPass = mapIsValid && evalResult.overallPass;
+
+		if (debugForcePass) {
+			mapIsValid = true;
+			overallPass = true;
+		}
 
 		resultVerdict.textContent = overallPass ? "Map Passed!" : "Map Failed";
 		resultVerdict.className = overallPass ? "pass" : "fail";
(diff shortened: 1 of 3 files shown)


game/web/src/main.ts
```diff
@@ -957,6 +965,98 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
+			}
+		}
+
+		// Debug: reset a single row to pending and re-run its reveal with a forced result.
+		// Cancels any in-flight animations (main reveal or a prior replay) before starting.
+		function debugReplayRow(row: HTMLElement, newPassed: boolean): void {
+			for (const t of activeTimeouts) clearTimeout(t);
+			activeTimeouts = [];
+			row.dataset["passed"] = String(newPassed);
+			row.dataset["finalized"] = "false";
```
cgruber, line 1027: **Stale `{ once: true }` listener on `btnRevealSkip` after `debugReplayRow` fires**

`debugReplayRow` sets `skipClickHandler` to a new cancel lambda, but it doesn't remove the
main-reveal `skipHandler` that was registered on `btnRevealSkip` as `{ once: true }` (line ~1207
in the new file). If a debug replay is triggered while the main reveal is still in flight (i.e.,
before the main reveal's chain has finished and hidden `resultRevealControls`), two things can
happen when the user then clicks the Skip button:

1. The `once` listener fires `skipHandler`, which iterates all `rowElements` and calls
   `finalizeRow` on them — including the row currently mid-replay. Because `debugReplayRow` resets
   `dataset["finalized"]` to `"false"`, that row will be finalized by the stale handler to
   whatever `dataset["passed"]` says. This is largely benign (the `finalized` guard prevents
   double-work on already-done rows), but it's unintentional: the old skipHandler snaps the
   replay row to final state outside the replay's own lifecycle.

2. `skipHandler` also calls `resultRevealControls.style.display = "none"`, hiding the Skip button
   mid-replay — the replay has no way to re-show it once it finishes naturally.

**Fix**: at the top of `debugReplayRow`, remove the old listener before installing the new cancel
hook:

```ts
if (skipClickHandler) {
    btnRevealSkip?.removeEventListener("click", skipClickHandler);
    skipClickHandler = null;
}
```

(The same pattern already used at the top of `showResultScreen` itself, line 885.)

This is only reachable in `?debug` mode, so it's not a production correctness issue, but it's
worth tightening since the PR explicitly targets reliable debug replay.

game/web/src/main.ts
```diff
@@ -957,6 +965,98 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
+					const gender = democode.length >= 2 ? democode.slice(-1) : "";
+					clipName = (gender === "m" || gender === "f")
+						? `${type}-${audioState}-${gender}`
+						: `${type}-${audioState}`;
+				} else if (type === "judge") {
+					clipName = `judge-${audioState}`;
+				} else {
+					// party
+					clipName = passed ? "party-approve" : "party-disapprove";
+				}
```
cgruber, line 1011: **Animation reset: `row.style.animation = "none"` before reflow is correct, but the 50 ms gap makes the reflow stale**

The reset sequence is:
```ts
row.style.animation = "none";
void row.offsetHeight;            // forces reflow synchronously
// ... then 50 ms later:
row.style.animation = `criterionReveal ${ROW_FADE_MS}ms ease forwards`;
```

Setting `animation` to `"none"` and immediately forcing a reflow is the standard browser trick to
restart a CSS animation. That part is fine. However, the reflow happens synchronously *before* the
50 ms `setTimeout`, while the animation re-trigger happens *inside* that timeout. In practice the
50 ms gap is long enough that no second reflow is needed, so this works. The only mild concern is
that the `void offsetHeight` on line 112 is now doing double duty as both "wait for pending style
flush" and "force layout" — and the real restart actually depends on the timer. If 0 ms were
sufficient, the 50 ms could be 0; the 50 ms is likely there to let the `rc-pending` class and
`opacity: ""` reset paint before the animation fires, which is reasonable.

No action required — just flagging that this is intentional and correct, not accidentally stale.

game/web/src/main.ts
cgruber: Fixed. Added the same guard that showResultScreen uses at its top: removes the stale once-listener from btnRevealSkip and nulls skipClickHandler before clearing activeTimeouts. Pushed.

game/web/src/main.ts
cgruber: No change. The `void row.offsetHeight` reflow happens synchronously before the 50ms setTimeout fires, so the browser processes the `animation: none` state before the new animation name is assigned. The 50ms timeout is what gives the animation a chance to re-trigger after the reflow — this is the standard CSS animation reset pattern.

PR #236, opened 2026-05-18
feat(GAME-073): deferred verdict banner + star count reveal

cgruber:
## Prerequisites
- [x] Parent PR merged: #235

## Summary
- On result screen open, verdict and subtitle start hidden (opacity 0); the banner is revealed progressively — failure banner appears on the first failing required criterion, success banner + star count appear after all rows resolve (or on skip/reduced-motion)
- `computeStarCount` fixed to follow DESIGN-001: returns 0 if any required criterion fails; otherwise 1 (base star) + count of optional criteria passed
- Stars rendered as filled/empty ★/☆ spans with a scale-in animation (disabled in `prefers-reduced-motion`) inside new `#result-stars` element
- `play("tada"/"womp-womp")` wired for completion audio — clips not yet generated (POST-MERGE)
- `syncOverallVerdict` (debug replay) extended to also refresh star display
- e2e: three new GAME-073 tests; existing tests updated with explicit `page.emulateMedia({ reducedMotion:"reduce" })` calls in helpers (required because `playwright.config reducedMotion:'reduce'` is ignored in Bazel-sandboxed Chromium)

## Validation performed
- [x] N/A — kubectl dry-run (not infra)
- [x] N/A — sops decrypt (not infra)
- TypeScript typecheck passes
- Full e2e test suite passes locally (135 tests)

- Web game only (result screen UX)

- OPTIONAL POST-MERGE: generate tada/womp-womp audio clips (~1–2s each, −16 LUFS) per GAME-073 AC; add to INVENTORY.md; preload with scenario

## Tickets addressed
- see #236 (GAME-073)

## Rollback
- Revert this commit; self-contained to `showResultScreen`, `computeStarCount`, `syncOverallVerdict`, styles, and index.html

### DIFF
--- a/game/web/e2e/sprint4.spec.ts
+++ b/game/web/e2e/sprint4.spec.ts
@@ -156,3 +156,54 @@ test("GAME-052: final state has correct criteria count after skip (regression)",
   const badges = page.locator(".result-criterion .rc-badge");
   await expect(badges).toHaveCount(count);
 });
+
+// ─── GAME-073: deferred success banner + star count reveal ───────────────────
+
+// Reduced-motion path: failure banner shows immediately (map unsolved = failing).
+test("GAME-073: failure banner shown immediately in reduced-motion mode", async ({ page }) => {
+  await loadEditor(page);
+  await openResultScreen(page); // already emulates reducedMotion:'reduce'
+  // Unsolved tutorial-002 has required-failing criteria.
+  await expect(page.locator("#result-verdict")).toHaveText("Map Failed");
+  await expect(page.locator("#result-stars")).toBeHidden();
+});
+
+// Reduced-motion path: success banner + stars shown immediately on a force-win.
+test("GAME-073: success banner and stars shown immediately in reduced-motion mode (force-win)", async ({ page }) => {
+  await page.goto("/?s=tutorial-002&debug");
+  await page.emulateMedia({ reducedMotion: "reduce" });
+  const skip = page.locator("#btn-intro-skip");
+  await expect(skip).toBeVisible({ timeout: 10_000 });
+  await skip.click();
+  await expect(page.locator("path.hex").first()).toBeVisible({ timeout: 10_000 });
+
+  await page.locator("#btn-debug-win").click();
+  await expect(page.locator("#result-screen")).toBeVisible({ timeout: 10_000 });
+
+  await expect(page.locator("#result-verdict")).toHaveText("Map Passed!");
+  await expect(page.locator("#result-stars")).toBeVisible();
+  // tutorial-002: 2 required + 1 optional → max 2 stars; force-win passes all → 2 stars
+  const starCount = await page.locator("#result-stars .result-star").count();
+  expect(starCount).toBe(2);
+});
+
+// Animated path: verdict starts empty, skip button reveals it with stars.
+test("GAME-073: skip reveals verdict and stars (animated mode, force-win)", async ({ page }) => {
+  await page.goto("/?s=tutorial-002&debug");
+  // Do NOT emulate reduced-motion — use animated path.
+  const skip = page.locator("#btn-intro-skip");
+  await expect(skip).toBeVisible({ timeout: 10_000 });
+  await skip.click();
+  await expect(page.locator("path.hex").first()).toBeVisible({ timeout: 10_000 });
+
+  await page.locator("#btn-debug-win").click();
+  await expect(page.locator("#result-screen")).toBeVisible({ timeout: 10_000 });
+
+  // Verdict starts empty before any row resolves.
+  await expect(page.locator("#result-verdict")).toHaveText("");
+
+  // Skip → all rows finalized → success banner appears.
+  await page.locator("#btn-reveal-skip").click();
+  await expect(page.locator("#result-verdict")).toHaveText("Map Passed!");
+  await expect(page.locator("#result-stars")).toBeVisible();
+});
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -95,6 +95,7 @@ const resultSubtitle = document.getElementById("result-subtitle") as HTMLElement
 const resultCriteriaList = document.getElementById("result-criteria-list") as HTMLElement | null;
 const btnKeepDrawing = document.getElementById("btn-keep-drawing") as HTMLButtonElement | null;
 const btnNextScenario = document.getElementById("btn-next-scenario") as HTMLButtonElement | null;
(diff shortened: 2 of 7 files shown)


game/web/styles.css
```diff
@@ -571,6 +572,32 @@ body {
...
   margin-top: -10px;
+  transition: opacity 0.3s ease;
+}
+
+#result-stars {
+  display: flex;
+  justify-content: center;
+  gap: 6px;
+  margin-top: -4px;
+}
```
cgruber, line 583: **Bug: `.hidden` class does not hide `#result-stars` — missing CSS rule**

`index.html` initialises the element as `<div id="result-stars" class="hidden">`, and the JS uses `classList.add("hidden")` / `classList.remove("hidden")` throughout.  However the stylesheet has no `#result-stars.hidden { display: none; }` rule.  The only rule for this element is the new `#result-stars { display: flex; … }` block, which always wins.

Result: on every page load the element is rendered as an empty flex container (0 height, but participates in layout and passes `toBeVisible` if Playwright checks bounding box), and the "failure banner shows immediately" test assertion `toBeHidden()` relies on the element being hidden — it will pass only because the element is empty, not because the class works.

Fix: add `#result-stars.hidden { display: none; }` to `styles.css`, following the same pattern used for `#main-menu.hidden`, `#result-screen.hidden`, etc.

game/web/src/main.ts
```diff
@@ -990,6 +1031,8 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 			const neutral = row.querySelector<HTMLElement>(".rc-char-neutral");
 			const verdict = row.querySelector<HTMLElement>(".rc-char-verdict");
 			if (neutral && verdict) transitionCharSlot(neutral, verdict);
+			// GAME-073: reveal failure banner on first required-criterion fail.
+			if (!passed && required) revealVerdict(false, 0);
```
cgruber, line 1041: **Audio overlap on first required-criterion fail**

`revealVerdict(false, 0)` (line 1035) fires `play("womp-womp")` before the `if (withAudio)` block (line 1036) fires the character's disapprove clip (e.g. `commissioner-disapprove-f`).  On the first required-criterion fail in the animated path both clips fire simultaneously on the same turn.

The `verdictShown` guard prevents this from repeating for subsequent rows, but for row 0 the two calls to `play()` overlap.  Depending on the audio queue implementation this may cause the banner stinger to interrupt the character clip or vice versa.

Consider moving the `revealVerdict` call to after the `withAudio` block, or inserting a short delay before the banner audio (e.g. let the character clip finish first).  At minimum, document the intended behaviour if simultaneous is acceptable.

game/web/src/main.ts
```diff
@@ -973,6 +976,44 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
+
+		function revealVerdict(pass: boolean, starCount: number): void {
+			if (verdictShown) return;
+			verdictShown = true;
+			resultVerdict!.textContent = pass ? "Map Passed!" : "Map Failed";
+			resultVerdict!.className = pass ? "pass" : "fail";
+			resultVerdict!.style.opacity = "1";
+			resultSubtitle!.textContent = pass
+				? "All required criteria met."
+				: mapIsValid
```
cgruber, line 1002: **Accessibility: star rating has no text alternative for screen readers**

Every `<span>` star gets `aria-hidden="true"`, so screen readers see the parent `#result-stars` container as empty.  A user who hears "Map Passed!" has no way to know the star count.

The `#result-verdict` element also has no `aria-live` region, so the deferred reveal (opacity transition) may not be announced at all in the animated path — the DOM text changes but the element was already rendered when the screen was first read.

Suggested fixes:
1. Add a visually-hidden text child to `#result-stars` that states the count: e.g. `<span class="sr-only">3 out of 3 stars</span>` (update it in `renderStars`).
2. Mark `#result-verdict` (or a wrapper) as `aria-live="polite"` so the deferred text update is announced after the reveal.

game/web/e2e/sprint4.spec.ts
```diff
@@ -156,3 +156,53 @@ test("GAME-052: final state has correct criteria count after skip (regression)",
...
+  await skip.click();
+  await expect(page.locator("path.hex").first()).toBeVisible({ timeout: 10_000 });
+
+  await page.locator("#btn-debug-win").click();
+  await expect(page.locator("#result-screen")).toBeVisible({ timeout: 10_000 });
+
+  await expect(page.locator("#result-verdict")).toHaveText("Map Passed!");
+  await expect(page.locator("#result-stars")).toBeVisible();
+  const starCount = await page.locator("#result-stars .result-star").count();
+  expect(starCount).toBeGreaterThan(0);
```
cgruber: **Test gap: `computeStarCount` formula not verified precisely**

The new formula (`1 + optional_passed`) is a meaningful change from the old `required_passed` count, but the e2e tests only assert `starCount > 0` (sprint4.spec.ts line ~89).  No test pins the actual count, e.g. "scenario with 2 optional criteria all passing yields 3 stars" or "1 optional failing yields 2 stars."

A unit test or a tighter assertion (e.g. `expect(starCount).toBe(expectedStars)`) would catch regressions in the formula without adding a lot of test complexity.  Even adding a `.result-star.filled` count assertion on the force-win test (where `maxStars` is known from the scenario definition) would help.

game/web/styles.css
cgruber: Fixed. Added `#result-stars.hidden { display: none; }` to styles.css. The element correctly hides when the class is set.

game/web/src/main.ts
cgruber: Fixed. When `pass=false`, womp-womp is now delayed by 400ms (`setTimeout(() => play("womp-womp"), 400)`) so the criterion's disapprove clip has time to start before the verdict audio fires.

game/web/src/main.ts
cgruber: Fixed. Added `aria-live="polite"` to `#result-verdict` in index.html so deferred text changes are announced. `renderStars` now sets `aria-label="N of M stars"` on the container each time it renders; individual star spans keep `aria-hidden="true"`. Static `aria-label="Star rating"` on the HTML element is overwritten by renderStars so the count is always current.

game/web/e2e/sprint4.spec.ts
cgruber: Fixed. tutorial-002 has 1 optional criterion (sc-compactness), so force-win yields `1 + 1 = 2` stars. The assertion is now `expect(starCount).toBe(2)` with a comment explaining the formula.

PR #238, opened 2026-05-19
chore(planning): S13 sprint plan, 11 new tickets, VRA research

cgruber:
## Summary

- Closes out sprint S11 and S12 in the roadmap and adds Sprint 13 (tutorial system + VRA scenarios) with Sprint 14 sketched (information density + terrain)
- Adds 11 new design/game tickets: DESIGN-012–015, GAME-075–081
- Resolves GAME-073 (result-screen stars & success banner), expands DESIGN-008 with a full terrain model, and adds VRA legal/political-landscape research docs
- No game code, build files, or test files are touched

## Affected machines / services

N/A — docs-only change

## Validation performed

- [x] `jj status` confirms only thoughts/shared/ files changed; no game code touched
- [x] Ticket IDs checked against TICKETS.md and Resolved section — no collisions
- [x] TICKETS.md Open section lists DESIGN-012–015, GAME-075–081; Resolved section includes GAME-073
- [x] Sprint roadmap updated: S11 complete, S12 complete, S13 planned, S14 sketched
- [x] Compressed sprint roadmap matches .md counterpart

## Deployment steps

N/A — no deployable artifacts

## Tickets addressed

- Resolves GAME-073 (result screen stars & success banner)
- New tickets: DESIGN-012, DESIGN-013, DESIGN-014, DESIGN-015, GAME-075, GAME-076, GAME-077, GAME-078, GAME-079, GAME-080, GAME-081

## Rollback

N/A — reverting this PR has no production impact; ticket files and docs can be restored from git history if needed

PR #239, opened 2026-05-19
feat(GAME-080): live per-district demographic stat under district buttons

cgruber:
## Prerequisites
- [x] Parent PR merged: #238

## Summary
- Adds a live per-district demographic stat beneath each district button on scenarios that have a `majority_minority` criterion (GAME-080 Phase 1)
- Stat is derived automatically from the existing criterion's `group_filter` and `min_eligible_share` — no new scenario spec fields required
- Districts meeting or exceeding the threshold are highlighted green (`.met` CSS class); all others show in muted text
- Stats update on every paint stroke via the existing `updateUI()` subscription

**Approach:** scan `scenario.success_criteria` for a `majority_minority` criterion at load time; call the newly-exported `computeDistrictGroupShares` in `updateUI()` to get per-district shares on each render; pass as an optional `demoStat` param to `renderDistrictButtons`.

## Validation performed
- [x] `bazel build //game/web:app_typecheck_lib` — TypeScript type check passes
- [x] `bazel test //game/web:e2e_test` — all 140 tests pass (5 new GAME-080 tests added)
- [x] Verified on scenario-005 (Valle Verde): stat shows "N% Latino" under each of 5 district buttons; painting the valley into D1 shows 1 `.met` district

## Deployment steps
- POST-MERGE: deploy to dev via `game/release.main.kts -- prepare && ./release.main.kts -- deploy --env dev`

## Affected components (detail)
- `game/web/src/simulation/evaluate.ts` — exported `computeDistrictGroupShares`, added `DistrictDemoStat` interface and `groupFilterLabel` helper
- `game/web/src/render/panels.ts` — `renderDistrictButtons` gains optional `demoStat?: DistrictDemoStat` param
- `game/web/src/main.ts` — detects criterion at load, computes stat in `updateUI()`
- `game/web/styles.css` — `.district-demo-stat` and `.met` styles
- `game/web/e2e/sprint5.spec.ts` — 5 new e2e tests
- `game/web/BUILD.bazel` — registers `sprint5.spec.ts` in test data

## Test plan
- [x] `GAME-080: scenario-005 shows district-demo-stat under each district button`
- [x] `GAME-080: district-demo-stat shows Latino percentage text`
- [x] `GAME-080: painting the winning solution gives district 1 the .met class`
- [x] `GAME-080: non-majority-minority scenario shows no demographic stat`
- [x] `GAME-080: stat count matches district button count`

## Tickets addressed
- see GAME-080

## Rollback
- Revert this PR; no data migration needed

### DIFF
--- a/game/web/src/main.ts
+++ b/game/web/src/main.ts
@@ -9,7 +9,7 @@
  */
 
 import { loadScenario } from "./model/loader.js";
-import type { Scenario, CriterionId, CharacterType } from "./model/scenario.js";
+import type { Scenario, CriterionId, CharacterType, Criterion } from "./model/scenario.js";
 import { type MapRenderer, type ViewMode, SvgMapRenderer } from "./render/mapRenderer.js";
 import {
 	renderDistrictButtons,
@@ -18,7 +18,14 @@ import {
 	renderValidityPanel,
 } from "./render/panels.js";
 import { createGameStore } from "./store/gameStore.js";
-import { evaluateCriteria, isMapSubmittable, type CriterionResult } from "./simulation/evaluate.js";
+import {
+	evaluateCriteria,
+	isMapSubmittable,
+	computeDistrictGroupShares,
+	groupFilterLabel,
+	type CriterionResult,
+	type DistrictDemoStat,
+} from "./simulation/evaluate.js";
 import { computeValidityStats } from "./simulation/validity.js";
 import {
 	loadProgress,
@@ -456,6 +463,12 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		return;
 	}
 
+	// GAME-080: detect majority_minority criterion for live district demographic stat
+	type MMCriterion = Extract<Criterion, { type: "majority_minority" }>;
+	const majorityMinorityCriterion = scenario.success_criteria
+		.map((sc) => sc.criterion)
+		.find((c): c is MMCriterion => c.type === "majority_minority");
+
 	// ── Preload audio clips (GAME-062) ───────────────────────────────────────
 	{
 		const clips: Record<string, string> = {};
@@ -566,9 +579,24 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		renderResults(resultsEl!, state, partyLabels);
 		renderValidityPanel(validityEl!, state, scenario.rules);
 		renderLegend(legendEl!, state.districtCount);
+
+		let demoStat: DistrictDemoStat | undefined;
+		if (majorityMinorityCriterion) {
+			demoStat = {
+				shares: computeDistrictGroupShares(
+					scenario.precincts,
+					state.assignments,
+					state.districtCount,
+					majorityMinorityCriterion.group_filter,
+				),
+				label: groupFilterLabel(majorityMinorityCriterion.group_filter),
+				threshold: majorityMinorityCriterion.min_eligible_share,
+			};
+		}
+
 		renderDistrictButtons(districtBtnsEl!, state.districtCount, state.activeDistrict, (id) => {
 			store.getState().setActiveDistrict(id);
-		});
+		}, demoStat);
 
 		btnUndo!.disabled = pastStates.length === 0;
 		btnRedo!.disabled = futureStates.length === 0;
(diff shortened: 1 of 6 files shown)


memory/feedback_shell_commands.md
```diff
@@ -0,0 +1,12 @@
+---
+name: Shell command preferences
+description: Use jq for JSON, Read tool with offset/limit for line ranges — avoid sed with line numbers
+type: feedback
+---
+
+Use `jq` for JSON evaluation, not python3 -c or other inline interpreters.
```
cgruber: **Scope mismatch: `memory/feedback_shell_commands.md` should not be in a feature PR**

This PR is GAME-080 (per-district demographic stat). The `memory/feedback_shell_commands.md` addition is a Claude agent auto-memory file with no connection to that feature. It should be split out to a separate `chore:` commit/PR, or simply excluded here.

Per the project workflow, memory file commits should not ride along on feature PRs — they're infrastructure-level changes that confuse the diff and could accidentally be reverted if this PR is rolled back.

Recommend: remove this file from the PR and land it separately.

game/web/src/render/panels.ts
```diff
@@ -55,17 +56,33 @@ export function renderDistrictButtons(
...
 		btn.style.color = "#fff";
 		btn.addEventListener("click", () => onSelect(i));
-		container.appendChild(btn);
+		wrapper.appendChild(btn);
+
+		if (demoStat) {
+			const share = demoStat.shares[i - 1] ?? 0;
+			const met = share >= demoStat.threshold;
+			const stat = document.createElement("div");
+			stat.className = `district-demo-stat${met ? " met" : ""}`;
```
cgruber, line 82: **Bug: display rounding and `.met` gate use different precision**

```ts
const met = share >= demoStat.threshold;   // raw float comparison
stat.textContent = `${Math.round(share * 100)}% ${demoStat.label}`;  // rounded display
```

A share of `0.497` with `threshold = 0.5` will display as **"50% Latino"** (because `Math.round(0.497 * 100) = 50`) but will NOT receive the `.met` green class (because `0.497 < 0.5`). The user sees "50%" in grey — visually contradictory.

Pick one source of truth:
- **Option A (recommended):** compare rounded values — `const pct = Math.round(share * 100); const met = pct >= Math.round(demoStat.threshold * 100);`
- **Option B:** use `Math.floor` for the display to never show a rounded-up number that's below threshold.

Either way, display and gate need to agree.

game/web/src/main.ts
```diff
@@ -456,6 +463,12 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
 		return;
 	}
 
+	// GAME-080: detect majority_minority criterion for live district demographic stat
+	type MMCriterion = { type: "majority_minority"; group_filter: GroupFilter; min_eligible_share: number; min_districts: number };
+	const majorityMinorityCriterion = scenario.success_criteria
```
cgruber, line 468: **Nit: inline `MMCriterion` type duplicates `Criterion` — use `Extract<>` instead**

```ts
type MMCriterion = { type: "majority_minority"; group_filter: GroupFilter; min_eligible_share: number; min_districts: number };
```

This duplicates the shape from `Criterion` in `scenario.ts`. If a field is added or renamed in the canonical union, this local alias won't catch it at compile time.

Prefer:
```ts
import type { ..., Criterion } from "./model/scenario.js";
type MMCriterion = Extract<Criterion, { type: "majority_minority" }>;
```

That way the type predicate stays automatically in lockstep with the source of truth.

memory/feedback_shell_commands.md
cgruber: Fixed. Removed `memory/feedback_shell_commands.md` from this commit via `jj restore --from @-`. The file already lives at the correct path in the Claude project memory directory — the repo-root copy was an accidental duplicate from the previous session.

game/web/src/render/panels.ts
cgruber: Fixed. Now compares rounded integer percentages on both sides: `const pct = Math.round(share * 100); const thresholdPct = Math.round(demoStat.threshold * 100); const met = pct >= thresholdPct;`. Display and highlight gate are now always consistent.

game/web/src/main.ts
cgruber: Fixed. Changed to `type MMCriterion = Extract<Criterion, { type: "majority_minority" }>;`. Now stays in lockstep with `scenario.ts` automatically — any upstream field rename will produce a compile error here.

PR #241, opened 2026-05-19
feat(GAME-075): terrain schema + loader (PR1 of 4)

cgruber:
## Prerequisites
- [x] N/A

## Summary
- GAME-075 PR1 of 4: terrain schema + loader (no rendering, no BFS changes — those land in PR2 and PR3)
- Adds non-precinct terrain tiles (sea/lake/mountain), edge rivers, and precinct terrain annotations (coast/lakeside/riverside/foothill) to the scenario format per DESIGN-008
- `scenario.ts`: new `TerrainType`, `TerrainAnnotation`, `TerrainTile` types; extends `Scenario` with `terrain_tiles?`, `river_edges?`, `river_blocks_contiguity?`; extends schema `Precinct` with `terrain?` and `has_internal_lake?`
- `types.ts`: extends runtime `Precinct` with optional `passableNeighbors`, `terrain`, `has_internal_lake`; extends `GameState` with optional `terrainTiles` and `riverEdges`; new `TerrainTileRuntime` interface
- `loader.ts`: parses terrain fields; validates terrain-tile-precinct-overlap, lake-adjacent-to-sea, mountain-enclosure (BFS from outside bounding box), and river-edge precinct references
- `adapter.ts`: computes `passableNeighbors` (river-blocked edges nulled when `river_blocks_contiguity: true`), derives terrain annotations from adjacency, converts river edges from string IDs to integer index pairs, builds runtime terrain tiles with pixel centers
- `gameStore.ts`: threads `terrainTiles` and `riverEdges` from adapter into `GameState`
- Plan document at `thoughts/shared/plans/2026-05-18-game-075-terrain.md` (+ compressed)

## Design decision — `neighbors` vs `passableNeighbors`
`Precinct.neighbors[i]` is used by both rendering (boundary segment computation) and BFS contiguity. When a river blocks contiguity, the renderer must still know the two precincts are geometrically adjacent (to draw the river stroke), but BFS must not traverse the river edge. Solution: keep `neighbors` intact for geometry; add `passableNeighbors` for BFS. PR3 will switch the BFS in `validity.ts` to read `passableNeighbors`.

## Validation performed
- [x] `bazel test //game/web/src/model:loader_test //game/web/src/model:adapter_test //game/web/src/simulation:validity_test //game/web/src/simulation:election_test //game/web/src/simulation:evaluate_test` — all pass
- [x] `bazel test //game/web:e2e_test` — 140 e2e tests pass (no regression on existing scenarios)
- [x] 7 new terrain loader tests added (happy path, tile/precinct overlap, lake-adjacent-sea, mountain enclosure, river edge unknown precinct, terrain field parsing, unknown annotation)
- [x] 6 new terrain adapter tests added (terrain tile conversion, coast derivation, explicit override, river edge conversion, passableNeighbors with/without blocking)

## Deployment steps
- POST-MERGE: no deploy needed; PR2 (renderer) lands the visible terrain features. Existing scenarios are unaffected (all terrain fields optional).

## Tickets addressed
- see GAME-075 (PR1 of 4 — schema + loader; renderer/contiguity/scenario follow in PR2/3/4)

## Rollback
- Revert this PR; no data migration needed (all new fields optional and absent from existing scenarios)

### DIFF
--- a/game/web/src/model/adapter.ts
+++ b/game/web/src/model/adapter.ts
@@ -14,7 +14,7 @@
  */
 
 import type { Scenario } from "./scenario.js";
-import type { AssignmentMap, DistrictId, Precinct } from "./types.js";
+import type { AssignmentMap, DistrictId, Precinct, TerrainTileRuntime } from "./types.js";
 import { HEX_DIRECTIONS, hexToPixel } from "./hex-geometry.js";
 
 // vote_shares is Record<PartyId, number> with branded keys; cast to plain string map at runtime
@@ -24,6 +24,8 @@ export function scenarioToSpike(scenario: Scenario): {
 	precincts: Precinct[];
 	assignments: AssignmentMap;
 	districtCount: number;
+	terrainTiles: TerrainTileRuntime[];
+	riverEdges: [number, number][];
 } {
 	// Map scenario DistrictId (branded string) → spike DistrictId (1-based number)
 	const districtIndexMap = new Map<string, DistrictId>();
@@ -38,6 +40,31 @@ export function scenarioToSpike(scenario: Scenario): {
 		if ("q" in pos) posMap.set(`${pos.q},${pos.r}`, i);
 	});
 
+	// Build terrain tile position map for annotation derivation
+	const terrainPosMap = new Map<string, "sea" | "lake" | "mountain">();
+	for (const tile of scenario.terrain_tiles ?? []) {
+		terrainPosMap.set(`${tile.position.q},${tile.position.r}`, tile.type);
+	}
+
+	// Build ID→index map for river edge conversion (precincts are ordered by scenario.precincts array)
+	const idToIndex = new Map<string, number>();
+	scenario.precincts.forEach((pc, i) => idToIndex.set(pc.id, i));
+
+	// Build river edge set: "idxA,idxB" (both orderings) for fast passableNeighbors lookup
+	const riverPairs: [number, number][] = [];
+	const riverEdgeSet = new Set<string>();
+	for (const [aId, bId] of scenario.river_edges ?? []) {
+		const aIdx = idToIndex.get(aId);
+		const bIdx = idToIndex.get(bId);
+		if (aIdx !== undefined && bIdx !== undefined) {
+			riverPairs.push([aIdx, bIdx]);
+			riverEdgeSet.add(`${aIdx},${bIdx}`);
+			riverEdgeSet.add(`${bIdx},${aIdx}`);
+		}
+	}
+
+	const blocksContiguity = scenario.river_blocks_contiguity ?? false;
+
 	const precincts: Precinct[] = scenario.precincts.map((pc, i) => {
 		const pos = pc.position;
 		const q = "q" in pos ? (pos as { q: number; r: number }).q : 0;
@@ -50,6 +77,33 @@ export function scenarioToSpike(scenario: Scenario): {
 			const idx = posMap.get(`${q + dq},${r + dr}`);
 			return idx !== undefined ? idx : null;
 		});
+
+		// passableNeighbors: same as neighbors but river edges nulled when blocking
+		const passableNeighbors: (number | null)[] = neighbors.map((nbIdx) => {
+			if (!blocksContiguity || nbIdx === null) return nbIdx;
+			return riverEdgeSet.has(`${i},${nbIdx}`) ? null : nbIdx;
+		});
+
+		// Derive terrain annotation from adjacency (explicit value overrides derivation).
+		// Priority: explicit > coast (sea) > lakeside (lake) > foothill (mountain) > riverside (river edge).
+		// Terrain tiles are dominant features; rivers are edge features, so tiles take priority.
+		let terrain: Precinct["terrain"] = pc.terrain as Precinct["terrain"] | undefined;
+		if (terrain === undefined) {
+			let hasSea = false;
+			let hasLake = false;
+			let hasMountain = false;
+			for (const [dq, dr] of HEX_DIRECTIONS) {
+				const tileType = terrainPosMap.get(`${q + dq},${r + dr}`);
+				if (tileType === "sea") hasSea = true;
+				else if (tileType === "lake") hasLake = true;
+				else if (tileType === "mountain") hasMountain = true;
+			}
+			const isRiverside = riverPairs.some(([a, b]) => a === i || b === i);
+			if (hasSea) terrain = "coast";
+			else if (hasLake) terrain = "lakeside";
+			else if (hasMountain) terrain = "foothill";
+			else if (isRiverside) terrain = "riverside";
+		}
 
 		// Population-weighted vote shares (turnout ignored until Sprint 3)
 		// Map first scenario party → R, second → D (matches partyIdToKey in main.ts)
@@ -79,6 +133,7 @@ export function scenarioToSpike(scenario: Scenario): {
 			coord: { q, r },
 			center,
 			neighbors,
+			passableNeighbors,
 			population: pc.total_population,
 			partyShare,
 			previousResult: { winner, margin },
@@ -86,6 +141,8 @@ export function scenarioToSpike(scenario: Scenario): {
 		};
 		if (pc.name !== undefined) spikePrecinct.name = pc.name;
 		if (pc.county_id !== undefined) spikePrecinct.county_id = pc.county_id;
+		if (terrain !== undefined) spikePrecinct.terrain = terrain;
+		if (pc.has_internal_lake) spikePrecinct.has_internal_lake = true;
 		if (pc.demographic_groups.length > 1) {
 			spikePrecinct.groupShares = pc.demographic_groups.map((g) => {
 				const entry: { name: string; share: number; dimensions?: Record<string, string> } = {
@@ -108,5 +165,11 @@ export function scenarioToSpike(scenario: Scenario): {
 		assignments.set(i, spikeDistId);
 	});
 
-	return { precincts, assignments, districtCount: scenario.districts.length };
+	// Build runtime terrain tiles with pixel centers
+	const terrainTiles: TerrainTileRuntime[] = (scenario.terrain_tiles ?? []).map((tile) => ({
+		center: hexToPixel(tile.position.q, tile.position.r),
+		type: tile.type,
+	}));
+
+	return { precincts, assignments, districtCount: scenario.districts.length, terrainTiles, riverEdges: riverPairs };
 }
--- a/game/web/src/model/adapter_test.ts
+++ b/game/web/src/model/adapter_test.ts
@@ -244,4 +244,177 @@ test("scenarioToSpike: districtCount matches scenario.districts.length", () => {
 	assertEqual(districtCount, 3, "districtCount = 3");
 });
 
+// ─── Terrain ──────────────────────────────────────────────────────────────────
+
+test("scenarioToSpike: terrain tiles converted with pixel centers", () => {
+	const g = makeGroup({ [pid("r_party")]: 0.5, [pid("d_party")]: 0.5 });
+	const scenario = makeScenario({
+		parties: [PARTY_R, PARTY_D],
+		districts: [{ id: did("d1") }, { id: did("d2") }],
+		precincts: [makePrecinct(0, 0, [g], "d1")],
+		terrain_tiles: [
+			{ position: { q: 1, r: 0 }, type: "sea" },
+			{ position: { q: 0, r: 1 }, type: "mountain" },
+		],
+	});
+	const { terrainTiles } = scenarioToSpike(scenario);
+	assertEqual(terrainTiles.length, 2, "two terrain tiles");
+	assertEqual(terrainTiles[0]!.type, "sea", "first is sea");
+	assertEqual(terrainTiles[1]!.type, "mountain", "second is mountain");
+});
+
+test("scenarioToSpike: coast annotation derived from adjacency to sea tile", () => {
+	const g = makeGroup({ [pid("r_party")]: 0.5, [pid("d_party")]: 0.5 });
+	const scenario = makeScenario({
+		parties: [PARTY_R, PARTY_D],
+		districts: [{ id: did("d1") }, { id: did("d2") }],
+		precincts: [makePrecinct(0, 0, [g], "d1")],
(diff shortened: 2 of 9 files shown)


game/web/src/model/adapter.ts
```diff
@@ -51,6 +78,28 @@ export function scenarioToSpike(scenario: Scenario): {
...
+			return riverEdgeSet.has(`${i},${nbIdx}`) ? null : nbIdx;
+		});
+
+		// Derive terrain annotation from adjacency (explicit value overrides derivation)
+		let terrain: Precinct["terrain"] = pc.terrain as Precinct["terrain"] | undefined;
+		if (terrain === undefined) {
+			for (const [dq, dr] of HEX_DIRECTIONS) {
+				const tileType = terrainPosMap.get(`${q + dq},${r + dr}`);
+				if (tileType === "sea") { terrain = "coast"; break; }
+				if (tileType === "lake") { terrain = "lakeside"; break; }
```
cgruber: **Terrain derivation priority bug: lake can shadow sea**

Both the `sea` and `lake` branches `break` out of the loop, so the FIRST one found wins — but the stated priority is sea > lake. Concrete failure case:

- Precinct at `(0, 0)`
- Lake tile at `(1, 0)` — direction 0 (lower-right)
- Sea tile at `(-1, 0)` — direction 3 (upper-left)

The two tiles are distance-2 apart (not adjacent), so Validation 2 (lake-adjacent-sea) does NOT reject this. The loop visits direction 0 first, finds `lake`, sets `lakeside`, and breaks before ever seeing the sea tile. Result: `lakeside` instead of `coast`.

This is realistic geography (peninsula with inland lake) — it will trigger on real scenarios.

Fix options:
1. Don't `break` on lake; only `break` on sea. After the loop, if `terrain === undefined` was upgraded to `lakeside`, keep it; if upgraded to `coast`, keep it. (Requires reordering the conditions or scanning all 6 first.)
2. Scan all 6 neighbors once into a `Set<TerrainType>`, then apply explicit priority: sea → coast > lake → lakeside > mountain → foothill.

Option 2 is clearer and avoids the order-dependence entirely. It's also trivially extensible if more terrain types arrive.

Add a test case: precinct adjacent to both sea and lake tiles → expect `coast`.

game/web/src/model/adapter.ts
```diff
@@ -51,6 +78,28 @@ export function scenarioToSpike(scenario: Scenario): {
...
+				if (tileType === "mountain" && terrain === undefined) terrain = "foothill";
+			}
+			// riverside: precinct appears in any river edge pair
+			if (terrain === undefined || terrain === "foothill") {
+				const isRiverside = riverPairs.some(([a, b]) => a === i || b === i);
+				if (isRiverside) terrain = "riverside";
```
cgruber: **Confirm intended priority: foothill vs riverside**

The header docstring/PR critique stated priority is "explicit > sea/coast > lake/lakeside > mountain/foothill > riverside", but the code below does the opposite — riverside overrides foothill:

```ts
if (terrain === undefined || terrain === "foothill") {
  const isRiverside = riverPairs.some(([a, b]) => a === i || b === i);
  if (isRiverside) terrain = "riverside";
}
```

So a precinct adjacent to a mountain AND on a river edge ends up `riverside`, not `foothill`. Which is intended? If foothill should win, drop `|| terrain === "foothill"`. If riverside should win, update the design doc/plan so the priority statement reflects the code.

Either way, add a test that pins down the chosen behavior so it doesn't drift silently.

game/web/src/model/loader.ts
```diff
@@ -903,6 +963,116 @@ export function loadScenario(json: unknown): Scenario {
...
+        }
+      }
+    }
+  }
+
+  // Validate river_edges: both precinct IDs must exist
+  if (river_edges !== undefined) {
+    for (const [aId, bId] of river_edges) {
+      if (!precinctIds.has(aId)) {
+        throw new Error(`river_edges: precinct "${aId}" does not exist in precincts`);
```
cgruber, line 1091: **Missing validation: river_edges precincts must be geometrically adjacent**

This only checks that both precinct IDs exist. But if `[p1, p7]` are listed and they're not hex-adjacent, two silent failures cascade:

1. `passableNeighbors` is unaffected — `riverEdgeSet.has(\`${i},${nbIdx}\`)` only ever nulls slots where `nbIdx` is already a real neighbor. So `river_blocks_contiguity` does nothing for the non-adjacent pair.
2. PR2's renderer plans to find the shared edge via `neighbors[i] === idxB`. For non-adjacent precincts that match fails, so the river silently doesn't render.

Result: scenario authors get no signal that their river edge is broken.

Suggested fix (cheap to add here): after building precincts in the adapter (or here in the loader, by computing neighbors per the hex_axial position set), verify each `[a, b]` in `river_edges` is in each other's neighbor set. Throw if not.

If you'd rather defer, file a follow-up ticket and add a TODO comment so it doesn't get lost.

game/web/src/model/loader.ts
```diff
@@ -903,6 +963,116 @@ export function loadScenario(json: unknown): Scenario {
...
+    }
+
+    // Build terrain position map for lake/sea adjacency check
+    const tileTypeMap = new Map<string, string>();
+    for (const tile of terrain_tiles) {
+      const key = `${tile.position.q},${tile.position.r}`;
+      // Terrain Validation 1: tile must not overlap a precinct position
+      if (precinctPosSet.has(key)) {
+        throw new Error(
+          `Terrain validation: terrain_tile at (${tile.position.q},${tile.position.r}) overlaps precinct position`
```
cgruber, line 996: **Custom geometry + terrain_tiles silently accepted, but unsupported**

The plan says: "For `custom` geometry: terrain tiles are not supported in v1 (spec is hex-grid only)." But the loader doesn't enforce this.

Downstream consequences when a scenario combines `geometry: { type: "custom" }` with `terrain_tiles`:

- `loader.ts:972` and `loader.ts:1014/1049`: the `"q" in pos` guard silently skips all custom-geometry precincts, so tile-overlap and mountain-enclosure validations never run for them.
- `adapter.ts:62-65`: `const q = "q" in pos ? ... : 0; const r = "q" in pos ? ... : 0;` — every custom precinct's adjacency lookup uses `(0, 0)` as its position, producing garbage annotation derivation.

Cheap fix: at the top of the terrain validation block in loader.ts, if `geometry.type === "custom"` and any of `terrain_tiles`, `river_edges`, or precincts with explicit terrain are present, throw a clear "terrain not supported with custom geometry" error.

(The explicit `precinct.terrain` field on custom-geometry precincts is a separate question — author may want to permit that since it doesn't require adjacency lookup. If so, gate only the adjacency-derived path.)

game/web/src/model/adapter_test.ts
```diff
@@ -244,4 +244,100 @@ test("scenarioToSpike: districtCount matches scenario.districts.length", () => {
...
+		river_edges: [[p1.id, p2.id]],
+		river_blocks_contiguity: true,
+	});
+	const { precincts } = scenarioToSpike(scenario);
+	// neighbors retains 1 for geometry; passableNeighbors nulls it
+	assertEqual(precincts[0]!.neighbors[0], 1, "neighbors[0] still 1 (geometry preserved)");
+	assertNull(precincts[0]!.passableNeighbors![0], "passableNeighbors[0] nulled (river blocks)");
+});
+
 summarize();
```
cgruber, line 420: **Test coverage gaps for derivation logic**

The PR adds explicit-annotation parsing tests and validation rejection tests, but several derivation paths in `adapter.ts` are untested:

1. **Sea+lake adjacency** — pinpoints the priority bug noted above.
2. **Foothill derivation** — precinct with a mountain neighbor (no sea/lake) should derive `foothill`.
3. **Riverside derivation** — precinct in a `river_edges` pair (with no adjacent terrain tile) should derive `riverside`.
4. **Foothill + riverside interaction** — pins down the priority question raised in another comment.
5. **`has_internal_lake` round-trip** — verify the boolean flows from scenario → spike Precinct (already parsed in loader_test.ts but not exercised through `scenarioToSpike`).

Without these, regressions on derivation rules will pass CI unnoticed.

game/web/src/model/adapter.ts
```diff
@@ -51,6 +78,28 @@ export function scenarioToSpike(scenario: Scenario): {
...
 		});
 
+		// passableNeighbors: same as neighbors but river edges nulled when blocking
+		const passableNeighbors: (number | null)[] = neighbors.map((nbIdx) => {
+			if (!blocksContiguity || nbIdx === null) return nbIdx;
+			return riverEdgeSet.has(`${i},${nbIdx}`) ? null : nbIdx;
+		});
+
+		// Derive terrain annotation from adjacency (explicit value overrides derivation)
+		let terrain: Precinct["terrain"] = pc.terrain as Precinct["terrain"] | undefined;
```
cgruber, line 90: **Critique summary (PR1 of 4, GAME-075)**

Verdict: **needs changes** — primarily one real correctness bug, plus two missing validations and gaps in derivation test coverage.

Inline issues posted:

1. **Sea-vs-lake derivation priority bug** (adapter.ts:93) — order-dependent due to `break` on lake; will trigger on realistic geographies (peninsula with inland lake).
2. **Foothill vs riverside priority** (adapter.ts:99) — code says riverside > foothill, prompt/spec note says foothill > riverside; need author intent confirmed and a pinning test.
3. **River edges not validated as geometrically adjacent** (loader.ts:1068) — only checks ID existence; non-adjacent pairs silently fail both `passableNeighbors` and PR2 rendering.
4. **Custom geometry + terrain_tiles silently accepted** (loader.ts:982) — plan says unsupported; loader doesn't reject; adapter falls back to q=r=0 for all custom precincts, producing garbage adjacency.
5. **Derivation test gaps** (adapter_test.ts) — no tests for sea+lake adjacency, foothill, riverside, or has_internal_lake adapter round-trip.

Things that look good:
- Plan/scope compliance: PR1 stays in schema + loader; BFS in validity.ts/evaluate.ts still reads `p.neighbors`, so existing flows are untouched.
- `passableNeighbors` is optional in the runtime type and the new field is unused outside this PR until PR3.
- Mountain enclosure BFS bounding box (minQ-2 .. maxQ+2 with one-layer slack) is sufficient for circumnavigation in axial hex space.
- HEX_DIRS duplication in loader.ts: justified per plan to avoid the hex-geometry dep on loader_lib; acceptable.
- River edge ordering — `precinctIds` is built before the `river_edges` existence check; correct.
- Validation 1 (overlap), 2 (lake-adj-sea), 3 (mountain enclosure) logic is correct for the hex_axial happy path.
- `pc.terrain as Precinct["terrain"]` cast is structurally safe (schema TerrainAnnotation == runtime union); only mild type-drift risk.

The headline fix is #1 (sea/lake priority) — that's a real bug, not theoretical. #2 is a quick design clarification. #3 and #4 can be addressed here or deferred with explicit TODOs + follow-up tickets.

game/web/src/model/adapter.ts
cgruber: Fixed. Refactored the derivation in adapter.ts to scan all 6 neighbors first (setting boolean `hasSea`/`hasLake`/`hasMountain` flags), then apply explicit priority: `coast > lakeside > foothill > riverside`. Iteration order no longer matters. Added test `sea takes priority over lake in derivation (coast wins)` placing sea and lake on opposite sides of a precinct to verify the fix.

game/web/src/model/adapter.ts
cgruber: Pinned to **foothill > riverside**. Rationale: terrain tiles are dominant geographic features; rivers are edge features. Mountain adjacency more strongly characterizes a precinct's geography than a single river edge. The full priority is now `explicit > coast > lakeside > foothill > riverside`, with an inline comment documenting it. Added test `foothill takes priority over riverside` to lock this in.

Note: this only affects the singular `terrain` annotation field used for the precinct's primary visual treatment. The river_edges data itself is still consumed independently by the river rendering layer (PR2), so a foothill precinct that's also on a river will still get its river stroke drawn — it just won't be classified as "riverside" in the terrain field.

game/web/src/model/loader.ts
cgruber: Fixed. Added geometric adjacency validation in loader.ts: after the existence check, the loader now computes `(bPos.q - aPos.q, bPos.r - aPos.r)` and verifies it matches one of the 6 HEX_DIRS vectors. Non-adjacent pairs are rejected with a descriptive error. Added test `rejects river edge between non-adjacent precincts` with precincts at (0,0) and (5,5).

game/web/src/model/loader.ts
cgruber: Fixed. Loader now rejects both `terrain_tiles` and `river_edges` when `geometry.type === "custom"` with an explicit "custom geometry is not supported in v1" error. This validation runs at the start of the terrain validation block so the existing check logic remains hex-axial only. Added two tests: `rejects terrain_tiles when geometry is custom` and `rejects river_edges when geometry is custom`.

game/web/src/model/adapter_test.ts
cgruber: Fixed. Added 5 new adapter tests covering the derivation logic gaps:

1. `sea takes priority over lake in derivation (coast wins)` — pins the priority bug fix from thread 1
2. `foothill annotation derived from mountain adjacency`
3. `riverside annotation derived when no terrain tile adjacency`
4. `foothill takes priority over riverside` — pins the priority decision from thread 2
5. `has_internal_lake round-trips through adapter`

Adapter test count: 17 → 22.

game/web/src/model/adapter.ts
cgruber: All 5 issues addressed:

1. **Sea-vs-lake derivation priority** — fixed (refactored to scan-then-apply priority); see thread 1 reply
2. **Foothill vs riverside priority** — pinned to foothill > riverside with rationale; see thread 2 reply
3. **River edge geometric adjacency** — validation added with adjacency check via HEX_DIRS; see thread 3 reply
4. **Custom geometry + terrain features** — both terrain_tiles and river_edges now rejected for custom geometry in v1; see thread 4 reply
5. **Test coverage** — 5 new adapter tests + 3 new loader tests; see thread 5 reply

Test counts after fixes:
- loader_test: 38 → 45 tests (7 + 3 new)
- adapter_test: 17 → 22 tests (6 + 5 new)

All tests pass. App builds. Ready for re-review.

PR #242, opened 2026-05-19
feat(GAME-075): terrain renderer — tiles, rivers, edge strokes (PR2 of 4)

cgruber:
## Prerequisites
- [x] Parent PR merged: #241

## Summary
- GAME-075 PR2 of 4: terrain renderer — visualizes the terrain data wired in by PR1 (#241). PR3 will switch the contiguity BFS to read `passableNeighbors`; PR4 will ship a demo scenario with terrain.
- Adds 3 new SVG groups to `SvgMapRenderer`: `terrainGroup` (below district borders — tile fills + glyphs), `terrainOverlayGroup` (above hex fills — coast/lakeside edge strokes + internal-lake ellipses), `riverGroup` (above county borders — blue river strokes along hex edges)
- 4 one-shot render helpers called from the constructor (terrain is immutable per scenario):
  - `renderTerrainTiles()` — sea (`#3a7fc1` + `∿`), lake (`#4dd0e1` + `∼`), mountain (`#6b7280` + `△`) with `pointer-events: none` so they're non-interactive
  - `renderRivers()` — derives segment coordinates from `state.riverEdges` pairs via `hexCorners()`; finds shared edge via `neighbors.findIndex(n => n === bIdx)`
  - `renderInternalLakes()` — aqua ellipse (~55%/45% of HEX_SIZE) centered in each `has_internal_lake: true` precinct
  - `renderTerrainEdges()` — sea-facing coast precincts get blue stroke, lake-facing lakeside precincts get aqua stroke on the specific edges that point at the terrain tile (looked up via axial coord)
- Stroke widths for rivers, terrain edges, and county borders now scale inversely with zoom alongside district/preview borders
- `types.ts`: `TerrainTileRuntime` gains `coord: HexCoord` (renderer needs axial position for edge adjacency)
- `hex-geometry.ts`: `HEX_SIZE` made `export const` so the renderer can size internal-lake ellipses against the canonical hex radius rather than recovering it from corner math

## Manual visual test plan
- Existing scenarios (no terrain data) load and render unchanged — verified by passing e2e suite (140 tests)
- Full visual verification with a real terrain scenario lands in PR4. Until then: schema and renderer can be exercised together by hand-editing a scenario JSON to add `terrain_tiles` / `river_edges` / `terrain` on precincts and reloading.

## Validation performed
- [x] `bazel test //game/web/src/...` — all 16 unit test targets pass
- [x] `bazel test //game/web:e2e_test` — 140 e2e tests pass (no regression on existing scenarios)
- [x] No regression on existing scenarios — they have no terrain fields, so all four render helpers early-return on empty data
- [ ] N/A — automated visual tests deferred to PR4 (no terrain scenario exists yet to assert against)

## Deployment steps
- POST-MERGE: no deploy needed; PR4 will ship a scenario that exercises this code path. Until then, terrain rendering is dormant.

## Tickets addressed
- see GAME-075 (PR2 of 4 — renderer; PR3=contiguity BFS, PR4=demo scenario)

## Rollback
- Revert this PR; PR1's schema/loader stays intact and terrain data is simply not rendered

### DIFF
--- a/game/web/src/render/mapRenderer.ts
+++ b/game/web/src/render/mapRenderer.ts
@@ -21,8 +21,8 @@
  */
 
 import * as d3 from "d3";
-import { hexCorners, mapBounds } from "../model/hex-geometry.js";
-import type { Precinct } from "../model/types.js";
+import { HEX_DIRECTIONS, HEX_SIZE, hexCorners, mapBounds } from "../model/hex-geometry.js";
+import type { Precinct, TerrainTileRuntime } from "../model/types.js";
 import { DISTRICT_COLORS, PARTY_LABELS } from "../model/types.js";
 import type { GameStore } from "../store/gameStore.js";
 
@@ -127,9 +127,12 @@ function computeCountySegments(precincts: Precinct[]): Segment[] {
 export class SvgMapRenderer implements MapRenderer {
 	private svg: SVGSel;
 	private zoomGroup: GSel;
+	private terrainGroup: GSel;
 	private countyBorderGroup: GSel;
 	private borderGroup: GSel;
 	private hexGroup: GSel;
+	private terrainOverlayGroup: GSel;
+	private riverGroup: GSel;
 	private previewBorderGroup: GSel;
 	private getState: () => GameStore;
 	private paintStroke: GameStore["paintStroke"];
@@ -196,6 +199,26 @@ export class SvgMapRenderer implements MapRenderer {
 	private countySegments: Segment[] = [];
 	private countyBordersVisible = false;
 
+	// Terrain styling (DESIGN-008): fills, glyphs, opacity for sea/lake/mountain tiles
+	private static readonly TERRAIN_FILL_SEA = "#3a7fc1";
+	private static readonly TERRAIN_FILL_LAKE = "#4dd0e1";
+	private static readonly TERRAIN_FILL_MOUNTAIN = "#6b7280";
+	private static readonly TERRAIN_GLYPH_SEA = "∿";
+	private static readonly TERRAIN_GLYPH_MOUNTAIN = "△";
+	private static readonly TERRAIN_GLYPH_COLOR = "rgba(255,255,255,0.45)";
+	private static readonly TERRAIN_GLYPH_FONT_SIZE = 22; // pixel size at 1× zoom
+	private static readonly RIVER_STROKE = "#38bdf8";
+	private static readonly RIVER_BASE_WIDTH = 2.5;
+	private static readonly RIVER_OPACITY = 0.7;
+	private static readonly INTERNAL_LAKE_FILL = "#4dd0e1";
+	private static readonly INTERNAL_LAKE_OPACITY = 0.55;
+	private static readonly INTERNAL_LAKE_RX_FACTOR = 0.55; // hex_size × factor = ellipse radius
+	private static readonly INTERNAL_LAKE_RY_FACTOR = 0.45;
+	private static readonly COAST_STROKE = "#3a7fc1";
+	private static readonly LAKESIDE_STROKE = "#4dd0e1";
+	private static readonly TERRAIN_EDGE_BASE_WIDTH = 3;
+	private static readonly TERRAIN_EDGE_OPACITY = 0.9;
+
 	// Keyboard precinct navigation state
 	private focusedPrecinctId: number | null = null;
 	private keyboardFocusPath: SVGPathElement | null = null;
@@ -213,12 +236,26 @@ export class SvgMapRenderer implements MapRenderer {
 		this.svg = d3.select(svgEl);
 
 		// Zoom group wraps all map layers so the single transform drives pan/zoom.
-		// Layer order (bottom → top inside zoomGroup): district borders, hexes, county borders, preview.
-		// County borders must sit above hexes so filled hex paths don't obscure them.
+		// Layer order (bottom → top inside zoomGroup):
+		//   terrainGroup        — non-precinct tiles (sea/lake/mountain), bottom
+		//   borderGroup         — committed district boundaries (solid white)
+		//   hexGroup            — precinct fills
+		//   terrainOverlayGroup — coast/lakeside edge strokes + internal lakes (above hex fills)
+		//   countyBorderGroup   — county boundary overlay (dashed gray; off by default)
+		//   riverGroup          — blue river strokes along hex edges (above precincts + county borders)
+		//   previewBorderGroup  — in-stroke boundary preview (top)
+		// Note on riverGroup placement: DESIGN-008 describes rivers as "above hex fills but
+		// below district outlines". In this renderer borderGroup is *below* hexGroup (district
+		// boundaries shine through semi-transparent hex fills), so the spec's two constraints
+		// are mutually exclusive. We choose "above hex fills" because rivers must be visible
+		// as natural geographic boundaries; they sit above district outlines as a result.
 		this.zoomGroup = this.svg.append("g").attr("class", "zoom-layer");
+		this.terrainGroup = this.zoomGroup.append("g").attr("class", "terrain");
 		this.borderGroup = this.zoomGroup.append("g").attr("class", "borders");
 		this.hexGroup = this.zoomGroup.append("g").attr("class", "hexes");
+		this.terrainOverlayGroup = this.zoomGroup.append("g").attr("class", "terrain-overlay");
 		this.countyBorderGroup = this.zoomGroup.append("g").attr("class", "county-borders");
+		this.riverGroup = this.zoomGroup.append("g").attr("class", "rivers");
 		this.previewBorderGroup = this.zoomGroup.append("g").attr("class", "preview-borders");
 
 		const pops = getState().precincts.map((p) => p.population);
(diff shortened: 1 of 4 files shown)


game/web/src/render/mapRenderer.ts
```diff
@@ -270,6 +309,162 @@ export class SvgMapRenderer implements MapRenderer {
...
+		const { precincts, riverEdges } = this.getState();
+		if (!riverEdges || riverEdges.length === 0) return;
+
+		const segments: Segment[] = [];
+		for (const [aIdx, bIdx] of riverEdges) {
+			const a = precincts[aIdx];
+			if (a === undefined) continue;
+			const edge = a.neighbors.findIndex((n) => n === bIdx);
+			if (edge < 0) continue;
+			const corners = hexCorners(a.center);
```
cgruber, line 395: nit (defense-in-depth): `findIndex(...) < 0` silently drops the river edge. Loader does validate that `river_edges` pairs are geometrically adjacent (loader.ts ~line 1104), so this branch is unreachable for valid scenarios. But if a future code path ever bypassed loader validation (test fixture, generator, custom geometry), a river would just vanish from the map with no signal.

Suggest a `console.warn` here — costs nothing for valid input, makes the rare bad-input case visible:

```ts
if (edge < 0) {
  console.warn(`renderRivers: river edge (${aIdx},${bIdx}) not in neighbors[] — skipping`);
  continue;
}
```

Not blocking.

game/web/src/render/mapRenderer.ts
```diff
@@ -196,6 +199,27 @@ export class SvgMapRenderer implements MapRenderer {
 	private countySegments: Segment[] = [];
 	private countyBordersVisible = false;
 
+	// Terrain styling (DESIGN-008): fills, glyphs, opacity for sea/lake/mountain tiles
+	private static readonly TERRAIN_FILL_SEA = "#3a7fc1";
+	private static readonly TERRAIN_FILL_LAKE = "#4dd0e1";
+	private static readonly TERRAIN_FILL_MOUNTAIN = "#6b7280";
+	private static readonly TERRAIN_GLYPH_SEA = "∿";
+	private static readonly TERRAIN_GLYPH_LAKE = "∼";
+	private static readonly TERRAIN_GLYPH_MOUNTAIN = "△";
```
cgruber, line 207: Lake glyph contrast: `TERRAIN_GLYPH_COLOR = rgba(255,255,255,0.45)` on top of `TERRAIN_FILL_LAKE = #4dd0e1` is very low contrast — white at 45% over light aqua blends to a near-invisible ribbon. Sea (`#3a7fc1`) and mountain (`#6b7280`) are fine because they're dark.

DESIGN-008 explicitly allows skipping the lake glyph: "Optional small wave or **none**" (line 82). Two reasonable options:

1. Drop the lake glyph (return `""` from `terrainGlyph("lake")`) — matches spec's "or none".
2. Use a darker glyph color specifically for lake (e.g. `rgba(20,60,80,0.45)` or a darker semi-transparent navy).

Sea reads fine with the white glyph; only the lake case is the issue. Not blocking but worth a tweak before PR4.

game/web/src/render/mapRenderer.ts
```diff
@@ -213,12 +237,21 @@ export class SvgMapRenderer implements MapRenderer {
...
+		//   countyBorderGroup   — county boundary overlay (dashed gray; off by default)
+		//   riverGroup          — blue river strokes along hex edges (above precincts + county borders)
+		//   previewBorderGroup  — in-stroke boundary preview (top)
 		this.zoomGroup = this.svg.append("g").attr("class", "zoom-layer");
+		this.terrainGroup = this.zoomGroup.append("g").attr("class", "terrain");
 		this.borderGroup = this.zoomGroup.append("g").attr("class", "borders");
 		this.hexGroup = this.zoomGroup.append("g").attr("class", "hexes");
+		this.terrainOverlayGroup = this.zoomGroup.append("g").attr("class", "terrain-overlay");
 		this.countyBorderGroup = this.zoomGroup.append("g").attr("class", "county-borders");
+		this.riverGroup = this.zoomGroup.append("g").attr("class", "rivers");
```
cgruber, line 258: Layer-order spec deviation (worth confirming, not necessarily a defect):

DESIGN-008 line 99-100 specifies rivers should be "above precinct hex fills but **below** district outlines." Here `riverGroup` is appended *after* `borderGroup` (district outlines), so rivers will render on top of the white district boundary anywhere a river edge coincides with a district boundary.

There's a structural reason this is hard to satisfy literally: `borderGroup` is below `hexGroup` in this codebase (the white outline is drawn first and "shows through" at hex edges where boundary stroke width > hex border width — see `renderBoundaries`). A river drawn between hexGroup and borderGroup would be hidden beneath the hex fills.

Probably the spec wording predates this layering technique. Current ordering (rivers on top) is the visually sensible compromise — but worth a sentence in the PR description (or an ADR) acknowledging the deviation so future readers don't try to "fix" it back to the literal spec.

game/web/src/render/mapRenderer.ts
```diff
@@ -231,6 +264,12 @@ export class SvgMapRenderer implements MapRenderer {
 		this.initBrushEvents();
 		this.initKeyboardNav();
 		this.initHoverEvents();
+
+		// One-shot terrain rendering — terrain is immutable per scenario.
+		this.renderTerrainTiles();
```
cgruber, line 273: ## Review: PR2 of GAME-075 — terrain renderer

**Verdict: LGTM with minor concerns** (none blocking).

### What I verified

- **Edge index correspondence**: `HEX_DIRECTIONS[i]` ↔ `corners[i]→corners[(i+1)%6]` ↔ `precinct.neighbors[i]` — confirmed against `hex-geometry.ts` docstring, `computeBoundarySegments`, and the adapter's neighbor construction. `renderTerrainEdges` is correct.
- **River edge lookup**: `findIndex` is safe because loader (loader.ts ~1078-1108) enforces geometric adjacency on `river_edges`. Posted a nit re: silent skip.
- **One-shot rendering**: scenario switching uses `window.location.assign(...)` (full page reload, main.ts ~227), so the renderer is constructed exactly once per page. No re-render hazard.
- **HEX_SIZE export visibility**: only one new consumer (mapRenderer for ellipse radii); test file has its own local copy. Safe.
- **TerrainTileRuntime construction sites**: only `adapter.ts:169`. The new `coord` field is correctly populated everywhere it's built.
- **Internal lake "no bridging"**: matches plan (`2026-05-18-game-075-terrain.md` line 106-108).
- **No foothill rendering**: matches plan ("deferred" / "may defer to post-S13").
- **No BFS or adapter logic changes** beyond the `coord` field add.
- **`pointer-events: none`** correctly set on all terrain elements.

### Inline comments

1. `renderRivers` silently drops un-findable edges — suggest `console.warn`.
2. Lake glyph contrast: white-on-aqua is low; spec allows dropping the lake glyph entirely.
3. `riverGroup` ordering vs DESIGN-008 spec line 99-100 — worth acknowledging the deviation; the spec wording predates the current borderGroup-below-hexGroup layering technique.

### Summary-level (not posted inline)

- No tests added — explicitly deferred to PR4 per plan, acceptable.
- ARIA / `aria-hidden="true"` on the new groups would be a nice cleanup since they're non-interactive; v1 acceptable.
- Class-field declaration order doesn't match the constructor append order — cosmetic.
- `data-join` key `String(i)` is fine given one-shot construction and immutable terrain.

Scope, plan compliance, type safety, and zoom-handler stroke-width updates all look correct.

game/web/src/render/mapRenderer.ts
cgruber: Fixed. Added `console.warn` on both the unknown-precinct-index path and the non-adjacent-neighbor path in `renderRivers`. The loader validates both invariants in PR1 so neither branch is reachable in a validated scenario, but warning defensively (rather than failing silently) is the right call if data ever bypasses the loader (e.g., scenario tests, hand-edited JSON in dev).

game/web/src/render/mapRenderer.ts
cgruber: Fixed. Dropped the lake glyph — DESIGN-008 explicitly lists it as optional ("Optional small wave or none"), and the aqua fill alone already distinguishes lakes from the darker sea. Avoids the low-contrast white-on-aqua issue entirely. Removed the now-unused `TERRAIN_GLYPH_LAKE` constant; `terrainGlyph("lake")` returns `""` with a comment explaining the omission.

game/web/src/render/mapRenderer.ts
cgruber: Acknowledged in the constructor layer-order comment. DESIGN-008 says "above hex fills but below district outlines" — but in this renderer borderGroup sits *below* hexGroup (district boundaries shine through semi-transparent hex fills, opacity 0.6 stroke vs 0.55–0.75 fill), so the spec's two constraints can't both hold. Chose "above hex fills" so rivers are visible as natural boundaries; they sit above district outlines as a consequence. Added an explanatory comment block in the layer-order section so the deviation is documented at the point of decision.

game/web/src/render/mapRenderer.ts
cgruber: All 3 actionable items addressed:

1. **River silent drop** — `console.warn` added to both `unknown-precinct` and `non-adjacent` paths in `renderRivers` (defense-in-depth; loader still validates the invariants in PR1).
2. **Lake glyph contrast** — lake glyph dropped per DESIGN-008's "Optional small wave or none". Constant `TERRAIN_GLYPH_LAKE` removed; helper documents the rationale.
3. **Layer-order deviation** — explanatory comment added to the constructor's layer-order block documenting why `riverGroup` sits above district outlines (borderGroup is below hexGroup in this renderer, making the spec's two constraints mutually exclusive).

Build green; all unit tests pass; 140 e2e tests pass. Ready for re-review.

PR #243, opened 2026-05-19
feat(GAME-075): contiguity BFS reads passableNeighbors (PR3 of 4)

cgruber:
## Prerequisites
- [x] Parent PR merged: #241
- [x] Parent PR merged: #242

## Summary
- GAME-075 PR3 of 4: contiguity BFS now reads `passableNeighbors` instead of `neighbors`. This activates the river-blocking behavior wired in by PR1 (#241) when a scenario sets `river_blocks_contiguity: true`.
- Single line change: `validity.ts:isContiguous` substitutes `p.passableNeighbors ?? p.neighbors` for the unconditional `p.neighbors` traversal. The `??` fallback preserves behavior for legacy/test precincts constructed without `passableNeighbors`.
- 3 new unit tests:
  - `river-blocked edge (passableNeighbors null) breaks BFS traversal` — two adjacent precincts with the shared edge nulled in `passableNeighbors` are reported non-contiguous.
  - `passableNeighbors absent falls back to neighbors` — legacy precincts still work.
  - `passableNeighbors preserves traversability for non-river edges` — a 3-precinct chain with only the middle edge blocked correctly splits into two components.

## Validation performed
- [x] `bazel test //game/web/src/simulation:validity_test` — all 14 tests pass (11 existing + 3 new)
- [x] `bazel test //game/web/src/...` — all unit test targets pass
- [x] `bazel test //game/web:e2e_test` — 140 e2e tests pass (no regression on existing scenarios; they don't set `river_blocks_contiguity`, and the existing scenarios have no river_edges to populate `passableNeighbors` differently from `neighbors`)

## Deployment steps
- POST-MERGE: no deploy needed; PR4 will ship a scenario that demonstrates river blocking. Until then, the behavior change is dormant.

## Tickets addressed
- see GAME-075 (PR3 of 4 — contiguity BFS; PR4=demo scenario)

## Rollback
- Revert this PR; BFS reverts to reading `neighbors` and river-blocking becomes dormant. Schema/loader/renderer remain intact.

### DIFF
--- a/game/web/src/simulation/validity.ts
+++ b/game/web/src/simulation/validity.ts
@@ -102,7 +102,11 @@ function isContiguous(
 		const curr = queue.shift()!;
 		const p = precincts[curr];
 		if (p === undefined) continue;
-		for (const nbId of p.neighbors) {
+		// passableNeighbors mirrors neighbors but nulls river-blocked edges when the
+		// scenario sets river_blocks_contiguity. Fall back to neighbors when absent
+		// (legacy/test precincts constructed without terrain data).
+		const traversable = p.passableNeighbors ?? p.neighbors;
+		for (const nbId of traversable) {
 			if (nbId !== null && inSet.has(nbId) && !visited.has(nbId)) {
 				visited.add(nbId);
 				queue.push(nbId);
--- a/game/web/src/simulation/validity_test.ts
+++ b/game/web/src/simulation/validity_test.ts
@@ -201,4 +201,74 @@ test("contiguity: two non-adjacent precincts in same district — not contiguous
   assertEqual(stats.contiguity!.get(2), true, "single-precinct district 2 is contiguous");
 });
 
+// ─── River-blocking contiguity (GAME-075) ─────────────────────────────────────
+
+/** Build a precinct with both neighbors and an explicit passableNeighbors override. */
+function makePrecinctWithBlocked(
+  id: number,
+  neighbors: (number | null)[],
+  passableNeighbors: (number | null)[],
+): Precinct {
+  return {
+    id,
+    coord: { q: 0, r: id },
+    center: { x: id * 10, y: 0 },
+    neighbors,
+    passableNeighbors,
+    population: 100,
+    partyShare: { R: 0.5, D: 0.5, L: 0, G: 0, I: 0 },
+    previousResult: { winner: "D", margin: 0 },
+    demographics: { male: 0.49, female: 0.49, nonbinary: 0.02 },
+  };
+}
+
+test("contiguity: river-blocked edge (passableNeighbors null) breaks BFS traversal", () => {
+  // P0 — P1 (adjacent geographically, river blocks)
+  // P0.neighbors[0] = 1 (river edge); P0.passableNeighbors[0] = null
+  // P1.neighbors[3] = 0; P1.passableNeighbors[3] = null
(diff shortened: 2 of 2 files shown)


game/web/src/simulation/validity.ts
```diff
@@ -102,7 +102,11 @@ function isContiguous(
 		const curr = queue.shift()!;
 		const p = precincts[curr];
 		if (p === undefined) continue;
-		for (const nbId of p.neighbors) {
+		// passableNeighbors mirrors neighbors but nulls river-blocked edges when the
+		// scenario sets river_blocks_contiguity. Fall back to neighbors when absent
+		// (legacy/test precincts constructed without terrain data).
+		const traversable = p.passableNeighbors ?? p.neighbors;
```
cgruber, line 108: ## LGTM — ship-ready

PR3 does exactly what the plan calls for: one-line BFS change + 3 focused unit tests. Reviewed against critique checklist:

1. **BFS correctness** — `p.passableNeighbors ?? p.neighbors` is right. `??` only triggers on null/undefined, so an empty `[]` would correctly block everything (not fall back). Adapter always populates `passableNeighbors` (`adapter.ts:82` maps over neighbors), so the fallback path is exercised only by legacy/test fixtures, which the inline comment correctly documents.
2. **Test coverage** — three cases cover blocking, fallback, and partial blocking on a 3-precinct chain. Asymmetric `passableNeighbors` (one direction passable, other blocked) isn't tested, but the adapter at `adapter.ts:53` stores river edges in both orderings, so asymmetry is precluded by construction. Non-issue.
3. **Side effects** — checked other `.neighbors` traversals:
   - `evaluate.ts:95` (`computeDistrictCompactness`) stays on `neighbors` — correct: compactness is a geometric measure, rivers don't change hex geometry.
   - `mapRenderer.ts` uses `neighbors` for edge/preview rendering — correct: also geometric.
   - No other simulation paths traverse neighbors.
4. **Plan compliance** — matches `2026-05-18-game-075-terrain.md` §PR3 ("one line change in isContiguous + tests") exactly.

### Two non-blocking nits (style only)

- **Test helper duplication**: `makePrecinctWithBlocked` duplicates most of `makePrecinct`'s boilerplate. Extending the existing helper with an optional `passableNeighbors?: (number | null)[]` parameter would be tidier. The new helper also drops the `population` parameter (hardcoded to 100) — slight asymmetry with the existing helper.
- **Safety pattern drift**: the existing contiguity tests at lines 180/189/199 use `assertNotNull(stats.contiguity, …)` before accessing the map; the three new tests skip that guard and use `stats.contiguity!.get(1)` directly. Tiny consistency thing — not worth blocking merge.

Verdict: approve. Tests pass on read-through; the change is minimal and the comment in `validity.ts` is clear about why the fallback exists.

cgruber: Thanks for the review. Acted on nit #2: added `assertNotNull(stats.contiguity, ...)` guards to all three new tests for consistency with the existing contiguity test pattern (validity_test.ts:180/189/199).

Skipped nit #1 (`makePrecinctWithBlocked` vs extending `makePrecinct`). Both shapes are defensible; the new helper has different defaults (omits the population parameter since the river-blocking tests don't exercise population balance). Keeping them parallel keeps each helper's intent local to its test suite. Happy to refactor if you prefer.

Issue #247, opened 2026-05-19
Terrain visual treatment refinement — thicker rivers, prominent coast/lakeside edges, foothill rendering


cgruber:
Terrain visual treatment refinement — thicker rivers, prominent coast/lakeside edges, foothill rendering

**Ticket:** [GAME-082](https://github.com/cgruber/redistricting-sim/blob/847ca917d7aee7bff6e8a2b41cb3d3011971b76a/thoughts/shared/tickets/GAME-082-terrain-visual-treatment-refinement.md)

PR #249, opened 2026-05-19
feat(GAME-082): smooth river curves via chain-building + d3.curveBasis (PR2)

cgruber:
## Prerequisites
- [x] Parent PR merged: #248

## Summary
- GAME-082 PR2: rivers now render as smooth meandering curves instead of straight hex-edge zigzags.
- Algorithm — `buildRiverChains(segs)` in `mapRenderer.ts`:
  1. Convert each river-edge precinct pair into its corner-pair pixel coordinates (via `hexCorners()`).
  2. Build a corner-key→segment-indices map. Corner keys are quantized to 2 decimals (`toFixed(2)`) to absorb any `Math.cos`/`Math.sin` float drift; adjacent edges share corner coordinates exactly because `hexCorners()` computes them identically from either side.
  3. Walk chains starting from degree-1 corners (natural chain endpoints). At any corner with degree ≠ 2 the walk terminates — this gracefully handles Y/T junctions where multiple river edges meet (each branch becomes its own chain, all meeting visually at the junction).
  4. Pass 2 emits any remaining unvisited segments as their own chains — covers closed-loop rivers and branches past a junction.
- Render — each chain becomes one `<path class="river-chain">` rendered via `d3.line().curve(d3.curveBasis)`. `curveBasis` approximates corner points without strictly passing through them, so the river meanders gently through the hex zigzag rather than tracing it.
- Side effect: scenario-010's 7 connected river edges now render as **1 path** instead of 7 separate lines.

## Junction handling
- Y/T-junction is split into multiple chains, each from an endpoint to the junction. They share a corner point so visually they meet at one place. Not optimal (the curves don't blend across the junction) but acceptable for v1 per the GAME-082 ticket. Future polish: emit a single multi-segment path with explicit M/L commands or sub-paths.
- Closed loop: walk passes through every segment and terminates back at the start when no unvisited candidate remains. Pass-2 emit picks it up.

## E2e test changes
- Updated `scenario-010 terrain: 7 river edges rendered along the bank boundary` → `scenario-010 terrain: river rendered as a single smoothed chain`. Asserts:
  - `path.river-chain` count = 1
  - `d` attribute starts with `M` (path move-to)
  - `d` attribute contains `C` (cubic bezier — confirms curveBasis output, not a fallback to straight lines)

## Validation performed
- [x] `bazel test //game/web:e2e_test` — 146 e2e tests pass (the updated scenario-010 river test passes against the new path-based rendering)
- [ ] N/A — `buildRiverChains` has no dedicated unit test; the render/ directory has no test target. The function is exercised end-to-end by the scenario-010 path assertion + manual eyeball after dev deploy. Adding a unit test would require either splitting the helper into a separate library file with its own BUILD entry or wiring up a new render-test target — out of scope for this PR.

- Static browser game only

## Deployment steps
- POST-MERGE: deploy to dev via `./game/release.main.kts -- prepare && ./game/release.main.kts -- deploy --env dev` to eyeball the new curved river rendering on scenario-010

## Tickets addressed
- see #247

## Rollback
- Revert this PR; rivers return to straight-line rendering. No data migration.

### DIFF
--- a/game/web/src/render/mapRenderer.ts
+++ b/game/web/src/render/mapRenderer.ts
@@ -54,6 +54,80 @@ type SVGSel = d3.Selection<SVGSVGElement, unknown, null, undefined>;
 // D3's append returns a Selection with parent type null when chained from select(element)
 type GSel = d3.Selection<SVGGElement, unknown, null, undefined>;
 type Segment = { x1: number; y1: number; x2: number; y2: number };
+type Point2D = [number, number];
+
+// ─── River chain building (GAME-082) ─────────────────────────────────────────
+
+/**
+ * Group river-edge corner-pair segments into connected chains so each chain
+ * can be rendered as a single smoothed SVG path rather than N separate lines.
+ *
+ * Two segments are considered connected when they share a corner (same pixel
+ * coordinates). Corners that appear in exactly 2 segments are interior to a
+ * chain; corners that appear in 1 segment are chain endpoints; corners with
+ * degree >= 3 are Y/T junctions where chains terminate (each branch becomes
+ * its own chain — visually they meet at the junction).
+ *
+ * Walking strategy: start from each endpoint (degree-1 corner) and follow
+ * unvisited segments through degree-2 corners. After all endpoint-rooted walks
+ * complete, any remaining unvisited segments form closed loops or branches at
+ * junctions — emit those as their own chains from an arbitrary starting point.
+ */
+function buildRiverChains(segs: [Point2D, Point2D][]): Point2D[][] {
+	// Quantize corner coordinates to handle hexCorners() float drift.
+	// 2 decimals = 0.01 px precision — well below any real edge mismatch.
+	const key = (p: Point2D): string => `${p[0].toFixed(2)},${p[1].toFixed(2)}`;
+
+	// corner-key → indices of segments touching that corner
+	const cornerToSegs = new Map<string, number[]>();
+	segs.forEach((seg, i) => {
+		for (const c of seg) {
+			const k = key(c);
+			if (!cornerToSegs.has(k)) cornerToSegs.set(k, []);
+			cornerToSegs.get(k)!.push(i);
+		}
+	});
+
+	const visited = new Set<number>();
+	const chains: Point2D[][] = [];
+
+	const walk = (startSeg: number, startCorner: Point2D): Point2D[] => {
+		const points: Point2D[] = [startCorner];
+		let segIdx = startSeg;
+		let corner = startCorner;
+		while (!visited.has(segIdx)) {
+			visited.add(segIdx);
+			const seg = segs[segIdx];
+			if (seg === undefined) break;
+			const other = key(seg[0]) === key(corner) ? seg[1] : seg[0];
+			points.push(other);
+			const candidates = (cornerToSegs.get(key(other)) ?? []).filter((s) => !visited.has(s));
+			if (candidates.length !== 1) break; // endpoint, junction, or dead end
+			segIdx = candidates[0]!;
+			corner = other;
+		}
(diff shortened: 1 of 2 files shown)


game/web/src/render/mapRenderer.ts
```diff
@@ -54,6 +54,80 @@ type SVGSel = d3.Selection<SVGSVGElement, unknown, null, undefined>;
...
+		const startCorner = key(seg[0]) === k ? seg[0] : seg[1];
+		chains.push(walk(segIdx, startCorner));
+	}
+	// Pass 2: any remaining unvisited segments belong to closed loops or live
+	// past a junction. Emit each as its own chain from an arbitrary start.
+	segs.forEach((seg, i) => {
+		if (visited.has(i)) return;
+		chains.push(walk(i, seg[0]));
+	});
+
```
cgruber, line 128: LGTM — chain-walker + curveBasis approach is sound.

Verified:

- **Chain-walker correctness**: For scenario-010's 7 connected edges, Pass-1 starts from one degree-1 endpoint and walks 7 segments → 8 corner points → one chain. Loop terminates safely (each iteration adds to `visited`; bounded by `segs.length`). `candidates.length !== 1` cleanly handles both endpoints (0) and junctions (≥2).
- **Quantization**: `toFixed(2)` (0.01 px) is well below any plausible float drift in `hexCorners()`. Safe across all reasonable axial coords since corner offsets from center are bounded by `HEX_SIZE`.
- **Render**: `fill="none"` present (essential — paths default to fill). `stroke-linejoin: round` is cheap insurance even though irrelevant for single-path curves. Zoom handler correctly updated to scale `path.river-chain`.
- **`Segment` type**: still in use for `EdgeSegment`, boundaries, county borders, preview borders — not dead code.
- **e2e**: assertions are appropriate for the single-branch river in scenario-010. `C` in the `d` attribute confirms curveBasis is actually engaged (vs. falling back to `L`).

Minor observations (non-blocking, possible follow-ups if other river topologies emerge):

1. **Pass-2 mid-chain start splits paths**: For a branch sitting between two junctions (rare — needs ≥2 junctions), Pass-2 picks an arbitrary unvisited segment and walks one direction only, so a single geometric branch could emit as 2 chains meeting mid-branch. Visually fine (they coincide), but each half is curve-fit independently so a tiny kink is possible at the split. Scenario-010 doesn't hit this; flag for the day we add a delta/braid.

2. **Single-segment chain ⇒ no `C`**: `d3.line().curve(d3.curveBasis)` on a 2-point input emits `L`, not `C`. The e2e assertion `d!.includes("C")` would fail if any future scenario added a one-edge river or a single-segment branch past a junction. Acceptable given current scenario set; just something to remember when authoring future scenarios.

3. **`lineGen(d) ?? ""`**: defensive — `d3.line` only returns null for empty arrays, and chains always have ≥1 point. Harmless.

Plan compliance ✓ (real curves via curveBasis, junctions terminate chains gracefully, no random naturalistic displacement).

cgruber: Thanks for the verification — chain-walker math, quantization tolerance, fill/linejoin, zoom scaling, and `Segment`-type still-in-use all reviewed in detail.

Acted on observation #2 (the only forward-compat issue worth touching): added an explicit comment to the e2e test noting that the `includes("C")` assertion is scenario-010-specific because curveBasis on 2-point chains emits `L` not `C`. Future scenarios with single-edge rivers will need their own assertion; this test stays scoped to scenario-010.

Observations #1 (Pass-2 mid-branch split) and #3 (`?? ""` defensive code) noted but not actioned — both are forward-compat shape rather than current correctness issues, and the Y/T-junction visual is already documented as "acceptable for v1" in the PR body.

PR #258, opened 2026-05-24
feat(GAME-084): loader behavioral split — parseScenario + validateScenarioComplete

cgruber:
## Summary

- **New `parseScenario(json): PartialScenario`** — structural parse only; tolerates absent gameplay fields (parties, districts, population, demographics); runs geometry/uniqueness/terrain invariants conditionally on data presence
- **New `validateScenarioComplete(partial): Scenario`** — completeness checks + full 13-invariant suite + auto-fill + assembly; call at game-start before gameplay
- **`loadScenario` unchanged externally** — refactored to `validateScenarioComplete(parseScenario(json))`; all existing callers continue to work
- **New `PartialPrecinct` + `PartialScenario` types** in `scenario.ts` — pipeline-stage types where `total_population`, `demographic_groups`, `parties`, `districts`, etc. are optional
- **25 new tests** in `loader_parse_test.ts` covering: terrain-only parse, completeness rejection, structural error detection at parse time, full-round-trip regression

This is AC 1 of GAME-084: separating load-time structural parsing from game-start completeness validation. The pipeline stages (terrain generator, population stage, demographics stage) will produce increasingly-complete `PartialScenario` JSON; `validateScenarioComplete` is called only when the game actually starts.

None — library change only; no new dependencies, no build changes other than one new test target.

## Validation performed

- [x] `bazel test //game/web/src/model:loader_parse_test` — 25/25 pass
- [x] `bazel test //game/web/src/model/...` — all 10 model test targets pass (includes 55 integration tests loading all 11 real scenario files)

N/A

## Tickets addressed

- GAME-084 AC 1 (see #258)

## Rollback

Revert commit

### DIFF
--- a/game/web/src/model/loader.ts
+++ b/game/web/src/model/loader.ts
@@ -1,13 +1,24 @@
 /**
  * Scenario JSON loader and validator.
  *
- * `loadScenario(json: unknown): Scenario` is the single entry point.
- * It validates all 13 spec invariants, auto-fills missing initial_district_id
- * on editable precincts, and returns a fully-typed Scenario.
+ * Three exported entry points:
+ *
+ *  `parseScenario(json: unknown): PartialScenario`
+ *    Structural parse: validates types and invariants that can be checked with
+ *    partial data (geometry consistency, unique IDs, context precincts, terrain).
+ *    Tolerates absent gameplay fields (parties, districts, population, demographics).
+ *    Use for pipeline intermediates and terrain-preview rendering.
+ *
+ *  `validateScenarioComplete(partial: PartialScenario): Scenario`
+ *    Completeness validation + full invariant check + auto-fill + assembly.
+ *    Throws if any gameplay field is absent. Call at game-start.
+ *
+ *  `loadScenario(json: unknown): Scenario`
+ *    Combined convenience: parseScenario + validateScenarioComplete.
+ *    Equivalent to the pre-GAME-084 single entry point; all callers continue to work.
  *
  * Every error message names the invariant number and identifies the offending
- * element (precinct id, group id, party id, etc.) so the caller knows where
- * the violation is.
+ * element so the caller knows where the violation is.
  */
 
 import type {
@@ -24,6 +35,8 @@ import type {
   Narrative,
   Party,
   PartyId,
+  PartialPrecinct,
+  PartialScenario,
   Precinct,
   PrecinctId,
   RegionSpec,
@@ -54,7 +67,7 @@ const HEX_DIRS: [number, number][] = [
 
 const EPSILON = 1e-6;
 
-// Parsed precinct before initial_district_id is resolved/auto-filled.
+// Internal type used by validateScenarioInvariants: all fields required (post-completeness-check).
 type RawPrecinct = Omit<Precinct, "initial_district_id"> & { initial_district_id?: DistrictId | null };
 
 // ─── Sub-parsers ─────────────────────────────────────────────────────────────
@@ -129,12 +142,11 @@ function parseDemographicGroup(raw: unknown, precinctId: string, idx: number): D
   return grp;
 }
 
-function parsePrecinct(raw: unknown, idx: number): RawPrecinct {
+function parsePrecinct(raw: unknown, idx: number): PartialPrecinct {
   const label = `precincts[${idx}]`;
   const r = requireObject(raw, label);
   const id = requireString(r["id"], `${label}.id`) as PrecinctId;
   const editable = requireBoolean(r["editable"], `${label}.editable`);
-  const total_population = requireNumber(r["total_population"], `${label}.total_population`);
 
   const posRaw = requireObject(r["position"], `${label}.position`);
   let position: Precinct["position"];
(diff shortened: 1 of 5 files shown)


game/web/src/model/loader.ts
```diff
@@ -1096,14 +1488,115 @@ export function loadScenario(json: unknown): Scenario {
...
+    if (pc.total_population === undefined) {
+      throw new Error(`completeness: precinct "${pc.id}" is missing total_population (required for gameplay)`);
+    }
+    if (pc.demographic_groups === undefined) {
+      throw new Error(`completeness: precinct "${pc.id}" is missing demographic_groups (required for gameplay)`);
+    }
+  }
+
+  // ── Full invariant check (all 13) ────────────────────────────────────────────
+  validateScenarioInvariants({
```
cgruber, line 1562: `validateScenarioComplete` calls `validateScenarioInvariants` (the full 13-invariant suite), but `validateStructural` — which already ran during `parseScenario` — also runs most of those same invariants conditionally when parties/districts/events/demographics are present.

In the `loadScenario(fullJson)` path, both passes fire on a fully-populated scenario, so invariants 1–9, 11, 12, and terrain are all checked twice. This isn't a correctness bug today, but it creates a maintenance trap: a future change to how one pass handles e.g. Invariant 7 (group_schema combos) or Invariant 3 (group refs in events) must be applied to **both** `validateStructural` and `validateScenarioInvariants`, or they'll silently diverge.

Suggested cleanup: in `validateScenarioComplete`, after the completeness checks, call `validateStructural(partial)` instead of `validateScenarioInvariants({…})`. Given all fields are now present, `validateStructural`'s conditionals all fire and it becomes equivalent. `validateScenarioInvariants` is then only called from `validateStructural` (or removed once `validateStructural` fully covers it). This eliminates the duplication without changing observable behavior.

Non-blocking, but worth tracking before the next pass on this file.

game/web/src/model/loader_parse_test.ts
```diff
@@ -0,0 +1,306 @@
...
+  delete s["rules"];
+  const partial = parseScenario(s);
+  assertThrows(() => validateScenarioComplete(partial), /rules/);
+});
+
+test("validateScenarioComplete: rejects partial missing narrative", () => {
+  const s = fullScenario() as Record<string, unknown>;
+  delete s["narrative"];
+  const partial = parseScenario(s);
+  assertThrows(() => validateScenarioComplete(partial), /narrative/);
```
cgruber, line 275: Three completeness-check branches in `validateScenarioComplete` have no test coverage:

1. **`events` missing**: line 1540–1542 (`partial.events === undefined` → throw) — no test exercises this path. The test at line 264 only deletes `rules`, not `events`.

2. **`success_criteria` missing**: line 1546–1548 (`partial.success_criteria === undefined` → throw) — similarly untested.

3. **`parties: []` (empty array, not undefined)**: line 1534 checks `partial.parties.length === 0`, but the only test for missing parties (line 226) uses a terrain-only partial that has `parties: undefined`. A scenario JSON with `"parties": []` would hit the `length === 0` branch exclusively and it isn't covered.

Suggested additions to `loader_parse_test.ts`:

```typescript
test("validateScenarioComplete: rejects partial missing events", () => {
  const s = fullScenario() as Record<string, unknown>;
  delete s["events"];
  const partial = parseScenario(s);
  assertThrows(() => validateScenarioComplete(partial), /events/i);
});

test("validateScenarioComplete: rejects partial missing success_criteria", () => {
  const s = fullScenario() as Record<string, unknown>;
  delete s["success_criteria"];
  const partial = parseScenario(s);
  assertThrows(() => validateScenarioComplete(partial), /success_criteria/i);
});

test("validateScenarioComplete: rejects partial with empty parties array", () => {
  const s = { ...fullScenario() as object, parties: [] };
  const partial = parseScenario(s);
  assertThrows(() => validateScenarioComplete(partial), /parties/i);
});
```

game/web/src/model/loader.ts
```diff
@@ -1049,44 +1059,426 @@ export function loadScenario(json: unknown): Scenario {
...
+function assembleScenario(
+  partial: PartialScenario,
+  parties: Party[],
+  districts: District[],
+  events: DemographicEvent[],
+  rules: ScenarioRules,
+  success_criteria: SuccessCriterion[],
+  narrative: Narrative,
+): Scenario {
+  const rawPrecincts = partial.precincts as RawPrecinct[];
```
cgruber, line 1460: The cast `partial.precincts as RawPrecinct[]` is safe in context: the completeness loop immediately above verifies that every precinct has `total_population` and `demographic_groups` before this line is reached, so at runtime the objects satisfy `RawPrecinct`'s shape.

However, TypeScript can't verify the invariant that makes the cast safe — it's only maintained by prose and call-order. If `assembleScenario` were ever called from a path that skips the completeness loop (e.g., a future internal shortcut), the cast would silently lie.

Consider a narrow type guard or a helper that converts `PartialPrecinct` → `RawPrecinct` after asserting both fields are defined, so the shape is verified in one place and the cast disappears. Non-blocking for this PR.

game/web/src/model/loader.ts
cgruber: Won't fix in this PR. The duplication is intentional for correctness safety. `validateScenarioInvariants` is the battle-tested full 13-invariant suite; collapsing `validateScenarioComplete` to rely solely on `validateStructural` would require a careful audit that the two are equivalent before any refactor. The double-firing is a no-op in practice — invariants are idempotent and data is small — so there's no correctness or performance issue. A cleanup PR to introduce a shared helper and eliminate the duplication is a good future improvement, but out of scope here.

game/web/src/model/loader_parse_test.ts
cgruber: Fixed. Added the three missing test branches to `loader_parse_test.ts`, all passing locally (`bazel test //game/web/src/model:loader_parse_test`, 28/28):

- **"validateScenarioComplete: rejects partial missing events"** — deletes `events` from `fullScenario()`, asserts throws matching `/events/`
- **"validateScenarioComplete: rejects partial missing success_criteria"** — deletes `success_criteria` from `fullScenario()`, asserts throws matching `/success_criteria/`
- **"validateScenarioComplete: rejects partial with empty parties array (length 0 branch)"** — uses `parties: []` with precincts that have no `demographic_groups` (so invariant 1 doesn't fire at parse time), asserts throws matching `/parties/i`

The empty-parties test needed precincts without demographic data so the party-ref invariant check at parse time doesn't pre-empt the completeness check in `validateScenarioComplete`.

game/web/src/model/loader.ts
cgruber: Won't fix in this PR. The cast `partial.precincts as RawPrecinct[]` is safe in context: the completeness loop immediately above it asserts `total_population !== undefined` and `demographic_groups !== undefined` on every precinct before the cast runs. A type-guard helper would be cleaner and eliminate the need for the assertion comment, but that refactor is out of scope for AC 1. Noted as a good future improvement.

Issue #259, opened 2026-05-25
Map generation pipeline — spec-driven, staged, single format


cgruber:
Map generation pipeline — spec-driven, staged, single format

**Ticket:** [GAME-084](https://github.com/cgruber/redistricting-sim/blob/ad29cef2c29bc7af4373f7e9ffef8c815dc2c658/thoughts/shared/tickets/GAME-084-map-generation-pipeline.md)

PR #264, opened 2026-05-31
feat(GAME-084): pipeline orchestrator + generate-scenario CLI

cgruber:
## Summary

Implements GAME-084 pipeline wiring: the `runPipeline` orchestrator and the `generate-scenario` CLI that ties all four generation stages together.

- **`pipeline-runner.ts`**: exports `runPipeline(spec: PipelineSpec): PartialScenario` — chains `generateTerrain` → `populateScenario` → `addDemographics`/`assignCounties` → `assembleScenario`; each stage is skipped when its spec section is absent; `county_labels` guarded by `.length > 0`
- **`pipeline-runner_test.ts`**: 10 tests — all stages present, each stage absent, `county_labels` absent vs empty array
- **`generate-scenario.ts`**: CLI reads YAML via `yaml` package, calls `runPipeline`, validates via `parseScenario + validateScenarioComplete`, writes JSON; `argv[2]`=spec path, `argv[3]`=optional output (defaults to replacing `.spec.yaml` → `.json`)
- **`tsconfig.cli.json`**: extends pipeline tsconfig; adds `"types": ["node"]` for the CLI
- **`BUILD.bazel`** (pipeline): adds `pipeline_runner_lib`, `pipeline_runner_typecheck_test`, `pipeline_runner_test_lib`, `pipeline_runner_test`, `tsconfig_cli`, `generate_scenario_lib`, `generate_scenario_typecheck_test`, `generate_scenario` (`js_binary`)
- **`model/BUILD.bazel`**: broadens `loader_lib` visibility to `//game/web:__subpackages__` (needed so the CLI can import `parseScenario`/`validateScenarioComplete`)

see #259 (GAME-084 wiring)

N/A — pipeline library + CLI tool only; no deployed services changed.

## Validation performed

- `bazel test //game/web/src/pipeline/...` — all 13 tests pass (12 existing + 1 new typecheck + 10 new unit tests)
- `bazel run //game/web/src/pipeline:generate_scenario -- <absolute>/scenario-002.spec.yaml /tmp/out.json` — produces valid scenario JSON: 91 precincts, 2 parties, 4 districts

## Deployment steps

N/A — library + CLI only; deployed as part of next game build.

## Tickets addressed

- GAME-084 (pipeline wiring)

## Rollback

Revert this PR. No data migrations or external state affected.

### DIFF
--- a/game/web/src/pipeline/generate-scenario.ts
+++ b/game/web/src/pipeline/generate-scenario.ts
@@ -1,0 +1,48 @@
+import * as fs from "node:fs";
+import * as path from "node:path";
+import { parse as parseYaml } from "yaml";
+import { runPipeline } from "./pipeline-runner.js";
+import { parseScenario, validateScenarioComplete } from "../model/loader.js";
+import type { PipelineSpec } from "./spec-types.js";
+
+const workspaceDir = process.env["BUILD_WORKSPACE_DIRECTORY"] ?? process.cwd();
+
+function resolvePath(p: string): string {
+  return path.isAbsolute(p) ? p : path.resolve(workspaceDir, p);
+}
+
+const args = process.argv.slice(2);
+if (args.length < 1) {
+  console.error("Usage: generate-scenario <spec-file.spec.yaml> [output.json]");
+  process.exit(1);
+}
+
+const specPath = resolvePath(args[0]!);
+
+let outputPath: string;
+if (args[1] !== undefined) {
+  outputPath = resolvePath(args[1]);
+} else {
+  const derived = specPath.replace(/\.spec\.yaml$/, ".json");
+  if (derived === specPath) {
+    console.error(`Error: spec path does not end in .spec.yaml; provide an explicit output path.`);
+    process.exit(1);
+  }
+  outputPath = derived;
+}
+
+const raw = fs.readFileSync(specPath, "utf-8");
+const spec = parseYaml(raw) as PipelineSpec;
+
+const partial = runPipeline(spec);
+
+const parsed = parseScenario(partial);
+const result = validateScenarioComplete(parsed);
+
+const outDir = path.dirname(outputPath);
+if (!fs.existsSync(outDir)) {
+  fs.mkdirSync(outDir, { recursive: true });
+}
+
+fs.writeFileSync(outputPath, JSON.stringify(result, null, 2) + "\n", "utf-8");
+console.log(`Written: ${outputPath}`);
--- a/game/web/src/model/BUILD.bazel
+++ b/game/web/src/model/BUILD.bazel
@@ -86,7 +86,7 @@ ts_project(
     tsconfig = ":tsconfig",
     declaration = True,
     deps = [":model_lib", ":runtime_types_lib"],
-    visibility = ["//game/web:__pkg__", "//game/web/src/model:__pkg__"],
+    visibility = ["//game/web:__subpackages__"],
 )
 
 # ── campaigns_lib: Campaign type + registry + last-played persistence ─────────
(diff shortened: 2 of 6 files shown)


game/web/src/pipeline/generate-scenario.ts
```diff
@@ -0,0 +1,31 @@
...
+const args = process.argv.slice(2);
+if (args.length < 1) {
+  console.error("Usage: generate-scenario <spec-file.spec.yaml> [output.json]");
+  process.exit(1);
+}
+
+const specPath = args[0]!;
+const outputPath = args[1] ?? specPath.replace(/\.spec\.yaml$/, ".json");
+
+const raw = fs.readFileSync(specPath, "utf-8");
```
cgruber, line 34: **`bazel run` cwd mismatch** — `fs.readFileSync(specPath, …)` resolves relative paths against `process.cwd()`, which under `bazel run` is the runfiles directory, not the workspace root. A path like `scenarios/foo.spec.yaml` will be sought inside the runfiles tree and will not be found.

The standard fix is to resolve relative paths against `process.env["BUILD_WORKSPACE_DIRECTORY"]` when that variable is set:

```ts
import * as path from "node:path";

function resolveArg(p: string): string {
  if (path.isAbsolute(p)) return p;
  const ws = process.env["BUILD_WORKSPACE_DIRECTORY"];
  return ws ? path.resolve(ws, p) : path.resolve(p);
}

const specPath = resolveArg(args[0]!);
const outputPath = resolveArg(args[1] ?? args[0]!.replace(/\.spec\.yaml$/, ".json"));
```

No other CLI binary in this repo uses `BUILD_WORKSPACE_DIRECTORY` yet, so this would set the convention. Without it, `bazel run //game/web/src/pipeline:generate_scenario -- specs/foo.spec.yaml` silently fails to find the file.

game/web/src/pipeline/generate-scenario.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { PipelineSpec } from "./spec-types.js";
+
+const args = process.argv.slice(2);
+if (args.length < 1) {
+  console.error("Usage: generate-scenario <spec-file.spec.yaml> [output.json]");
+  process.exit(1);
+}
+
+const specPath = args[0]!;
+const outputPath = args[1] ?? specPath.replace(/\.spec\.yaml$/, ".json");
```
cgruber: **Silent overwrite footgun** — if the caller passes a file whose name does not end in `.spec.yaml` (e.g. `foo.yaml`, `foo.json`), the `.replace(…)` returns the original path unchanged, so `outputPath === specPath`. Running `bazel run … -- foo.yaml` would overwrite `foo.yaml` with the generated JSON, silently deleting the spec.

A simple guard would prevent this:

```ts
const defaultOutput = specPath.replace(/\.spec\.yaml$/, ".json");
if (args[1] === undefined && defaultOutput === specPath) {
  console.error(
    `Error: input file must end in .spec.yaml, or an explicit output path must be provided.\n` +
    `Got: ${specPath}`
  );
  process.exit(1);
}
const outputPath = args[1] ?? defaultOutput;
```

game/web/src/pipeline/generate-scenario.ts
cgruber: Fixed. Now resolves relative paths against `process.env["BUILD_WORKSPACE_DIRECTORY"]` when set (bazel run sets this to the workspace root), falling back to `process.cwd()` for direct invocation. Verified: `bazel run … -- game/scenarios/scenario-002.spec.yaml` (relative path) now resolves correctly.

game/web/src/pipeline/generate-scenario.ts
cgruber: Fixed. Now detects when the default `replace()` was a no-op (input path unchanged) and exits with an error requiring the caller to supply an explicit output path.

PR #265, opened 2026-06-01
feat(GAME-087): terrain-aware population stage — suitability + settlement zones

cgruber:
## Summary

- Replaces flat-random `populateScenario` with a two-layer terrain-aware model
- **Layer 1 — Terrain suitability**: per-precinct multiplier from adjacency to lake/sea/mountain tiles and river edges (lakeside 1.4×, riverside 1.3×, coastal 0.9×, mountain-adjacent 0.5×); features compose multiplicatively; overridable via `terrain_weights` spec
- **Layer 2 — Settlement zones**: optional Gaussian population bumps (`peak × exp(-dist²/(2σ²))`); anchored to terrain features, exact hex coords, `center`, or cardinal directions
- `scenario-002.spec.yaml` updated with Clearwater City (anchor:center, peak:600, r:3) + East Mills (anchor:east, peak:200, r:1); peaks tuned conservatively to keep e2e districts within ±5% of mean (mean=36222; D1=+5%, D2=-5%, D3=-5%, D4=+4%)
- `scenario-002.json` regenerated; pop range 1356–2208, avg 1592

- N/A — pipeline tooling only; no deployed service changes

## Validation performed

- [x] `bazel test //game/web/src/pipeline:population_stage_test` — all tests pass (12 new tests covering all 10 ACs)
- [x] `bazel test //game/web/src/pipeline:...` — full pipeline suite green
- [x] e2e test suite passes with regenerated scenario-002.json (all 4 districts within ±5% of mean)
- [x] All 10 GAME-087 ACs met

## Deployment steps

- N/A — no deployment required; scenario JSON regenerated in-tree

## Tickets addressed

- see GAME-087 (ticket file at `thoughts/shared/tickets/GAME-087-terrain-aware-population-stage.md`; status→resolved in this PR)

## Rollback

- Revert this PR; scenario-002.json returns to flat-random population

### DIFF
--- a/game/web/src/pipeline/population-stage.ts
+++ b/game/web/src/pipeline/population-stage.ts
@@ -1,28 +1,242 @@
 /**
- * Stage 2 of the GAME-084 map generation pipeline: population stage.
+ * Stage 2 of the GAME-084 map generation pipeline: terrain-aware population stage.
  *
- * Takes a PartialScenario (from the terrain generator) and a PopulationSpec,
- * and assigns a deterministic integer total_population to every precinct.
+ * Assigns total_population to every precinct using a two-layer model:
  *
- * Semantics:
- *   - Always overwrites any existing total_population (spec is source of truth).
- *   - Population = base + prng.nextInt(-variance, variance), evaluated once per
- *     precinct in the order they appear in the precincts array.
- *   - Outputs are integers; no fractional populations.
+ *   Layer 1 — Terrain suitability (always on):
+ *     Each precinct gets a multiplicative suitability score based on adjacency to
+ *     terrain features (lake, sea, mountain tiles; river edges). Multiple features
+ *     compose multiplicatively. Default weights: lakeside 1.4×, riverside 1.3×,
+ *     coastal 0.9×, mountain-adjacent 0.5×. Overridable via terrain_weights.
+ *
+ *   Layer 2 — Settlement zones (optional spec):
+ *     Named settlements add a Gaussian population bump centered on an anchor
+ *     precinct. Anchor types: exact {q,r}, feature-based (lakeside/riverside/
+ *     coastal), center, or cardinal directions.
+ *
+ *   Final formula per precinct:
+ *     total_population = round(base × suitability + settlement_bump
+ *                              + prng.nextInt(-variance, variance) × suitability)
+ *
+ * Always overwrites existing total_population; input PartialScenario is not mutated.
  */
 
 import type { PartialScenario } from "../model/scenario.js";
-import type { PopulationSpec } from "./spec-types.js";
+import type {
+  HexPos,
+  PopulationSpec,
+  SettlementAnchor,
+  TerrainWeightsSpec,
+} from "./spec-types.js";
 import { makePrng } from "./prng.js";
+
+// ─── Defaults ─────────────────────────────────────────────────────────────────
+
+const DEFAULT_TERRAIN_WEIGHTS: Required<TerrainWeightsSpec> = {
+  lakeside: 1.4,
+  riverside: 1.3,
+  coastal: 0.9,
+  mountain_adjacent: 0.5,
+};
+
+// Flat-top axial hex direction vectors (same as terrain-generator.ts)
+const HEX_DIRS: [number, number][] = [
+  [1, 0], [0, 1], [-1, 1], [-1, 0], [0, -1], [1, -1],
+];
+
+// ─── Internal types ───────────────────────────────────────────────────────────
+
+interface TerrainContext {
+  lakeside: boolean;
+  riverside: boolean;
+  coastal: boolean;
+  mountain_adjacent: boolean;
+}
+
+// ─── Helpers ──────────────────────────────────────────────────────────────────
+
+function posKey(q: number, r: number): string {
+  return `${q},${r}`;
+}
+
+export function hexDist(a: HexPos, b: HexPos): number {
+  const dq = a.q - b.q;
+  const dr = a.r - b.r;
+  return (Math.abs(dq) + Math.abs(dr) + Math.abs(dq + dr)) / 2;
+}
+
+function buildTerrainContexts(partial: PartialScenario): Map<string, TerrainContext> {
+  const lake = new Set<string>();
+  const sea = new Set<string>();
+  const mountain = new Set<string>();
+
+  for (const tile of partial.terrain_tiles ?? []) {
+    const key = posKey(tile.position.q, tile.position.r);
+    if (tile.type === "lake") lake.add(key);
+    else if (tile.type === "sea") sea.add(key);
+    else if (tile.type === "mountain") mountain.add(key);
+  }
+
+  const riverside = new Set<string>();
+  for (const edge of partial.river_edges ?? []) {
+    riverside.add(edge[0]);
+    riverside.add(edge[1]);
+  }
+
+  const result = new Map<string, TerrainContext>();
+  for (const p of partial.precincts) {
+    const pos = p.position as HexPos;
+    let isLakeside = false;
+    let isCoastal = false;
+    let isMountainAdj = false;
+    for (const [dq, dr] of HEX_DIRS) {
+      const nk = posKey(pos.q + dq, pos.r + dr);
+      if (lake.has(nk)) isLakeside = true;
+      if (sea.has(nk)) isCoastal = true;
+      if (mountain.has(nk)) isMountainAdj = true;
+    }
+    result.set(p.id, {
+      lakeside: isLakeside,
+      riverside: riverside.has(p.id),
+      coastal: isCoastal,
+      mountain_adjacent: isMountainAdj,
+    });
+  }
+  return result;
+}
+
+function computeSuitability(ctx: TerrainContext, w: Required<TerrainWeightsSpec>): number {
+  let s = 1.0;
+  if (ctx.lakeside) s *= w.lakeside;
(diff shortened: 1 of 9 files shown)


game/web/src/pipeline/population-stage_test.ts
```diff
@@ -141,4 +148,290 @@ test("populateScenario: metadata fields are preserved unchanged", () => {
...
+  const flat = result.precincts.find(p => p.id === "p001")!;
+  const lakeside = result.precincts.find(p => p.id === "p002")!;
+  const ratio = lakeside.total_population! / flat.total_population!;
+  assertEqual(ratio > 1.35, true);
+  assertEqual(ratio < 1.45, true);
+});
+
+test("terrain: default mountain_adjacent weight is 0.5× flat (AC2 defaults)", () => {
+  const spec = basePopSpec({ base: 10000, variance: 10 });
+  const partial = makeTerrainPartial();
```
cgruber, line 299: Cardinal direction anchor tests are smoke-only — they only assert that the call doesn't throw. The `DIRECTION_SCORE` table is the most error-prone part of this file (sign flips between q/r are easy to get wrong), and the tests provide no behavioral coverage.

Suggest at least one test that verifies the correct precinct is actually selected, e.g.:

```ts
test("anchor: 'east' resolves to the precinct with maximum q", () => {
  const partial = makePartial(2); // hex_circle r=2, precincts at various (q,r)
  const spec = basePopSpec({ base: 1000, variance: 0,
    settlements: [{ anchor: "east", peak: 10000, radius: 0.5 }] });
  const result = populateScenario(partial, spec);
  // Find precinct with max q — should have bump ≈ peak
  const maxQ = Math.max(...partial.precincts.map(p => (p.position as { q: number; r: number }).q));
  const anchored = result.precincts.find(p => (p.position as { q: number; r: number }).q === maxQ)!;
  assertEqual(anchored.total_population! > 9000, true);
});
```

Without this, a sign flip in `DIRECTION_SCORE` (e.g. `east: -q`) would not be caught.

game/web/src/pipeline/population-stage_test.ts
```diff
@@ -141,4 +148,290 @@ test("populateScenario: metadata fields are preserved unchanged", () => {
...
+  // Custom lakeside = 2.0 → expect ~20000 vs default 1.4×~14000
+  assertEqual(lakeside.total_population! > 18000, true);
+  assertEqual(lakeside.total_population! > flat.total_population!, true);
+});
+
+test("terrain: no terrain features in partial — suitability 1.0, no throw", () => {
+  const partial = makePartial(2);
+  const spec = basePopSpec({ base: 1000, variance: 50 });
+  const result = populateScenario(partial, spec);
+  for (const p of result.precincts) {
```
cgruber, line 338: Feature anchor tests are also smoke-only (same pattern as cardinal directions). They verify no throw, but don't check that the bump actually lands on the correct precinct.

Consider asserting that the precinct matching the anchor type has higher population than those that don't. For example, for `anchor: "lakeside"`, the lakeside precinct should have a larger bump than any non-lakeside precinct in the same partial.

This would catch a regression where `resolveAnchor` picks the wrong candidate (e.g., if the `ctx.lakeside` / `ctx.riverside` branch is incorrectly structured).

game/web/src/pipeline/population-stage_test.ts
```diff
@@ -141,4 +148,290 @@ test("populateScenario: metadata fields are preserved unchanged", () => {
...
+//   p002 at (1,0): riverside
+//   p003 at (-2,0): flat (not adjacent to any terrain, not in any river edge)
+function makeRiverPartial(): PartialScenario {
+  return {
+    format_version: "1",
+    id: "river-test" as unknown as ScenarioId,
+    title: "River Test",
+    election_type: "congressional",
+    region: { id: "r1" as unknown as RegionId, name: "R" },
+    geometry: { type: "hex_axial" },
```
cgruber, line 188: The default weight ratio tests cover `lakeside` (1.4×) and `mountain_adjacent` (0.5×) but not `coastal` (0.9×) or `riverside` (1.3×). The coastal weight is the only one below 1.0 other than mountain, so it's worth pinning separately — a refactor that accidentally made it a multiplier > 1 would be a design regression.

Suggest adding:

```ts
test("terrain: default coastal weight is 0.9× flat (AC2 defaults)", () => {
  // Need a partial with a sea tile adjacent to one precinct
  ...
  const ratio = coastal.total_population! / flat.total_population!;
  assertEqual(ratio > 0.85, true);
  assertEqual(ratio < 0.95, true);
});
```

game/web/src/pipeline/population-stage_test.ts
```diff
@@ -141,4 +148,290 @@ test("populateScenario: metadata fields are preserved unchanged", () => {
...
+  const spec = basePopSpec({ base: 10000, variance: 10 });
+  const result = populateScenario(makeComposabilityPartial(), spec);
+  // p002: lakeside only (suitability = 1.4 → ~14000)
+  // p003: lakeside+riverside (suitability = 1.4 × 1.3 = 1.82 → ~18200)
+  const lakeside = result.precincts.find(p => p.id === "p002")!;
+  const lakeRiver = result.precincts.find(p => p.id === "p003")!;
+  assertEqual(lakeRiver.total_population! > lakeside.total_population!, true);
+});
+
+test("terrain: default lakeside weight is 1.4× flat (AC2 defaults)", () => {
```
cgruber, line 284: No error-path tests for `resolveAnchor`. Three `throw` sites exist in the implementation and none are exercised:

1. `{q,r}` anchor with no precinct at that position
2. Feature anchor (`lakeside`/`riverside`/`coastal`) on a map with no matching precincts
3. An unknown anchor string (guarded by `if (!scorer)` — currently reachable only from a runtime cast bypass)

Cases 1 and 2 are the realistic failure modes for a spec author who sets an `anchor` that doesn't match the generated terrain. Worth a test each:

```ts
test("resolveAnchor: throws if no precinct at exact HexPos anchor", () => {
  const spec = basePopSpec({ base: 1000, variance: 0,
    settlements: [{ anchor: { q: 999, r: 999 }, peak: 1000, radius: 1 }] });
  assertThrows(() => populateScenario(makePartial(1), spec));
});

test("resolveAnchor: throws if lakeside anchor but no lakeside precincts", () => {
  const spec = basePopSpec({ base: 1000, variance: 0,
    settlements: [{ anchor: "lakeside", peak: 1000, radius: 1 }] });
  // makePartial produces a flat map with no terrain
  assertThrows(() => populateScenario(makePartial(1), spec));
});
```

game/web/src/pipeline/population-stage.ts
```diff
@@ -1,28 +1,242 @@
...
+const DEFAULT_TERRAIN_WEIGHTS: Required<TerrainWeightsSpec> = {
+  lakeside: 1.4,
+  riverside: 1.3,
+  coastal: 0.9,
+  mountain_adjacent: 0.5,
+};
+
+// Flat-top axial hex direction vectors (same as terrain-generator.ts)
+const HEX_DIRS: [number, number][] = [
+  [1, 0], [0, 1], [-1, 1], [-1, 0], [0, -1], [1, -1],
```
cgruber, line 44: `HEX_DIRS` is duplicated verbatim from `terrain-generator.ts` (same comment and values). Not a bug today, but it's a drift risk — if the terrain generator ever switches to pointy-top orientation and updates its copy, this file won't track.

Consider extracting to a shared utility (e.g., `hex-utils.ts` or `hex-math.ts`) and importing in both places. `hexDist` is already exported from `population-stage.ts` and used in tests, so there's precedent for this kind of shared hex math.

game/web/src/pipeline/population-stage_test.ts
cgruber: Fixed. Added `settlement: east anchor places bump at max-q precinct` — computes `maxQ`, partitions precincts into at-max-q vs others, and asserts the max-q bucket has higher population than any other precinct.

game/web/src/pipeline/population-stage_test.ts
cgruber: Fixed. Added `settlement: lakeside anchor places bump at lakeside precinct` — uses `makeTerrainPartial()` (p002 is the only lakeside precinct), asserts p002 has higher total_population than all other precincts after a large-peak settlement.

game/web/src/pipeline/population-stage_test.ts
cgruber: Fixed. Added `makeCoastalPartial()` fixture (p001 flat, p002 adjacent to sea tile at (3,0)) and `terrain: default coastal weight is 0.9× flat (AC2 defaults)` — ratio test asserting coastal/flat ∈ (0.85, 0.95) with base=10000, variance=10.

game/web/src/pipeline/population-stage_test.ts
cgruber: Fixed. Added two error-path tests:
- `resolveAnchor: HexPos anchor throws when no precinct at position` — anchor {q:99,r:99} on a r=1 map
- `resolveAnchor: feature anchor throws when no matching precincts` — `coastal` anchor on makeTerrainPartial() which has no sea tiles

game/web/src/pipeline/population-stage.ts
cgruber: No change. `HEX_DIRS` duplication across pipeline stages is a known drift risk — noted as a follow-up refactor (extract to shared hex-utils) outside this ticket's scope.

PR #300, opened 2026-06-27
refactor(game): extract win/star verdict logic to simulation/verdict.ts (GAME-110)

cgruber:
## Summary

Extracts the pure win/star verdict logic out of the untested `main.ts` controller into a new,
unit-tested `simulation/verdict.ts`, and unifies the duplicated verdict-UI code paths. **Behavior
is unchanged** for all shipped scenarios — a structural/testability fix (review Theme 2; grok
recommended doing this **first** among the refactors). Resolves **GAME-110**.

- **`simulation/verdict.ts`** (new) — pure functions, verdict as a projection of evaluate +
  validity:
  - `computeStarCount` (moved verbatim)
  - `computeMaxStars` (extracted from the inline `maxStars` expression)
  - `buildValidityRows` (moved; made pure by taking the scenario criterion-type set as a
    parameter instead of closing over `scenario`)
- **`verdict_test.ts`** (new) — the 1-base + per-optional star rule, failed required/optional,
  invalid map = 0, and the `buildValidityRows` skip-when-equivalent + contiguity-row branches.
- **BUILD** — `verdict_lib` / `_typecheck_test` / `_test_lib` / `_test` mirroring the `evaluate_*`
  blocks; `verdict_lib` added to `app_typecheck_lib` + the esbuild `app` deps.
- **`main.ts`** — imports the three functions, deletes the inline defs, and extracts one
  `applyVerdictUI(pass, starCount, { subtitle })` that both `revealVerdict` and
  `syncOverallVerdict` call (each keeps its own guard / audio / subtitle), so the normal and
  debug-replay verdict paths can't diverge.

One intentional, verified no-op: `applyVerdictUI` hides stars on the fail path; the original
`revealVerdict` left them untouched on fail, but stars are already cleared+hidden during
evaluation setup before any reveal, so the effect is identical.

- The static web game (`game/web`). No services. Pure refactor + new tests.

## Validation performed

- [x] `bazel test //game/...` green — **41/41** (incl. e2e exercising the result-screen reveal, app typecheck, and the two new verdict targets).
- [x] New `verdict_test.ts` covers all star-count and validity-row branches.
- [x] Verdict UI dedup hand-verified behavior-preserving (incl. the stars-on-fail no-op).

## Deployment steps

## Tickets addressed

- Resolves GAME-110 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Pure refactor; reverting restores the inline implementation with no data/schema impact.


game/web/src/main.ts
```diff
@@ -1136,23 +1094,37 @@ const IS_DEBUG = (debugParam !== null && debugParam !== "off") ||
...
 
-		function revealVerdict(pass: boolean, starCount: number): void {
-			if (verdictShown) return;
-			verdictShown = true;
+		// GAME-110: single source of truth for the verdict banner — verdict
+		// text/class/opacity, subtitle, stars, and post-win button wiring. Both the
+		// normal reveal path (revealVerdict) and the debug-replay path
+		// (syncOverallVerdict) call this so they can't diverge. Callers own their
+		// own guards/audio and compute the subtitle string (pass it via opts).
+		function applyVerdictUI(pass: boolean, starCount: number, opts?: { subtitle?: string }): void {
```
cgruber: **The dedup's actual regression surface (DOM wiring) is verified by nothing (note, not a blocker)**

`verdict_test.ts` is good, but it only exercises the three pure functions in isolation. The change that carries real regression risk in this PR is the `applyVerdictUI` merge — and `verdict_test` structurally cannot reach it (no DOM). I checked the e2e suite: `scenarios.spec.ts` / `submit-on-invalid.spec.ts` assert `#result-verdict` text ("Map Passed!" / "Map Failed") and class on both paths, so the verdict-text branch of `applyVerdictUI` is covered. But **no test asserts**:
- the on-screen **star count** (`.result-star.filled`) — the `computeStarCount`/`renderStars` wiring;
- the **subtitle text** in any branch (`#result-subtitle`) — including the fail+invalid "structural issues" string that only `revealVerdict` produces;
- **stars hidden on fail**.

So divergence between the normal and debug-replay paths is now prevented *by construction* (shared call site), which is the right design — but it is not *measured*. This is acceptable for a behavior-preserving refactor; flagging so it's a conscious choice, not an assumed-covered one. A single assertion on `.result-star.filled` count + `#result-subtitle` text on one win path would close it.

Related latent footgun introduced here: `applyVerdictUI` takes `opts?: { subtitle?: string }` and falls back to `resultSubtitle.textContent = opts?.subtitle ?? ""`. Both current callers always pass an explicit subtitle, so behavior is preserved today. But the old inline code had **no** silent-blank path — every reveal computed a subtitle. A future third caller that forgets `subtitle` now silently renders a blank subtitle instead of failing. Consider making `subtitle` required (drop the `?`) so the type system keeps that invariant.

game/web/src/simulation/BUILD.bazel
```diff
@@ -121,6 +121,68 @@ js_test(
...
+ts_project(
+    name = "verdict_lib",
+    srcs = ["verdict.ts"],
+    tsconfig = ":tsconfig",
+    declaration = True,
+    deps = [
+        ":evaluate_lib",
+        ":validity_lib",
+        "//game/web/src/model:model_lib",
+        "//game/web/src/model:types_lib",
```
cgruber: **Unused `types_lib` dep on `verdict_lib` (nit)**

`verdict.ts` imports only from `../model/scenario.js` (`model_lib`), `./evaluate.js` (`evaluate_lib`), and `./validity.js` (`validity_lib`). It does **not** import anything from `model/types.ts` (`types_lib` — "spike/legacy types for the hex-grid renderer"). The `//game/web/src/model:types_lib` dep on both `verdict_lib` and `verdict_test_lib`/`verdict_test` was mirrored from `evaluate_lib`/`validity_lib` (which genuinely use it) and is dead here.

Why it matters: it's harmless to the build (no strict-deps enforcement), but it's exactly the kind of vestigial dep GAME-109 already calls out ("drop vestigial generator_lib dep"). Pruning it now keeps the new target honest.

Fix: drop `"//game/web/src/model:types_lib"` from `verdict_lib`, `verdict_test_lib` deps, and the `verdict_test` data list.

cgruber: Fixed — removed `//game/web/src/model:types_lib` from `verdict_lib`, `verdict_test_lib`, and `verdict_test`. Confirmed `verdict.ts` only needs `model_lib` (for `CriterionId` from `scenario.ts`), `evaluate_lib` (`CriterionResult`), and `validity_lib` (`ValidityStats`) — `types_lib` is the spike/legacy `types.ts`, genuinely unused here. `bazel test //game/...` still 41/41.

game/web/src/main.ts
cgruber: Partially addressed. Fixed the latent footgun: `applyVerdictUI`'s `subtitle` is now a required `string` param (removed the `opts?.subtitle ?? ""` silent-blank path); both callers always pass it.

On the coverage note: agreed it's real but I'm intentionally not expanding scope here. Divergence between the two verdict paths is now prevented by construction (single shared call site), and verdict_test locks the pure star/validity logic; adding e2e assertions for on-screen star count, subtitle text, and stars-hidden-on-fail is exactly the kind of result-screen coverage that belongs in GAME-109 (test backstops + coverage) rather than this behavior-preserving refactor. Noted there.

PR #301, opened 2026-06-28
refactor(loader): de-duplicate the dual scenario-validation passes (GAME-102)

cgruber:
## Summary

De-duplicates the invariant logic in `loader.ts` — the file that gates every scenario the game
loads. `validateStructural` (parse, partial-data) and `validateScenarioInvariants` (complete)
each re-implemented Invariants 1–9/11 and a near-byte-identical ~130-line terrain/river block; a
rule changed in one copy silently diverged the parse-time and gameplay paths (review Theme 1).
Resolves **GAME-102**.

**Approach — single definition per invariant, called from both phases** (decided after reading
the callers; the literal normalize-once collapse was rejected):

- Each duplicated check extracted to one module-level helper (`checkUniqueIds`,
  `checkGeometryAndNeighbors`, `checkVoteShareCompleteness`, `checkPopulationShareSums`,
  `checkDistrictRefs`, `checkGroupRefsInEvents`/`Criteria`, `checkGroupSchemaCompleteness`,
  `checkPartyRefsInEventsAndCriteria`, `validateTerrainAndRivers`, …).
- `validateStructural` calls them behind its existing `if (x !== undefined)` guards; the complete
  path calls them after the completeness gate. Helpers run in each function's original sequence,
  so **per-function throw order and every Error string are preserved verbatim**. Parse runs first
  on every production path, so `loadScenario`'s observable behavior is unchanged.
- **Why not the grok-literal normalize-once collapse:** the two-phase API is used separately by
  `generate-scenario.ts` (`parseScenario` then `validateScenarioComplete`) and is locked by
  `loader_parse_test.ts`, so collapsing the public surface would be a contract change, not a
  refactor.
- **Two inter-copy differences found + reconciled** (documented in the resolved ticket):
  Inv1/Inv6 per-group interleaving (per-group predicates preserve both paths' original order) and
  Inv7 loop structure (the shared helper adopts the canonical parse-first message; the only
  changed behavior is the first Inv7 message from `validateScenarioComplete` called *in isolation*
  — never via `loadScenario` — on a multi-violation input).

`loader.ts`: **1634 → 1468 lines (−166)**.

- The static web game (`game/web`) + the offline scenario pipeline (`generate-scenario.ts`, which calls the same loader). Behavior-preserving refactor.

## Validation performed

- [x] `bazel test //game/...` green — **41/41** (loader_test, loader_parse_test, loader_integration_test, loader_typecheck_test, e2e).
- [x] Added a **parse-vs-complete agreement** test block (Inv 11, Inv 8, terrain enclosure, river adjacency, the Inv1/Inv6 interleaving, Inv7 same-message) — locks single-definition behavior; mutation-verified that breaking either path fails it.
- [x] Error strings preserved verbatim (tests assert on `/Invariant N/` + terrain/river messages).

## Deployment steps

## Tickets addressed

- Resolves GAME-102 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Behavior-preserving refactor; reverting restores the duplicated implementation with no data/schema impact.


game/web/src/model/loader.ts
```diff
@@ -1120,358 +1280,70 @@ function validateStructural(s: PartialScenario): void {
...
-        const dq = bPos.q - aPos.q, dr = bPos.r - aPos.r;
-        const isAdjacent = HEX_DIRS.some(([ddq, ddr]) => ddq === dq && ddr === dr);
-        if (!isAdjacent) {
-          throw new Error(`river_edges: precincts "${aId}" (${aPos.q},${aPos.r}) and "${bId}" (${bPos.q},${bPos.r}) are not geometrically adjacent`);
-        }
-      }
-    }
-  }
+  // Terrain + river validation (conditional on terrain_tiles / river_edges)
+  validateTerrainAndRivers(geometry, precincts, precinctIds, terrain_tiles, river_edges);
```
cgruber, line 1346: **Independent equivalence audit — LGTM (changes-not-requested).**

I reviewed this as an adversarial equivalence audit against the pre-refactor file (`e9b04af`, the PR base), not a style pass. The "behavior-preserving" claim holds for **every production path**. What I verified:

**Runtime error strings are byte-identical.** I extracted all thrown `Invariant N:` / `Terrain validation:` / `river_edges:` templates from both the original and the refactored file and diffed the sets. The only deltas are *source identifiers* that produce identical runtime output:
- `${pc.id}` → `${precinctId}` (helper param bound to `pc.id`)
- Inv6 missing-party `${partyId}` → `${pid}` (parse copy already used `${pid}`; both bind the same party id)
- Inv6 sum `${sum}` → `${vsum}` (both are the same `Object.values(...).reduce(...)`)

No invariant message was dropped or altered. Inv2 precinct message is identical in both original copies (`...does not exist in districts`); only a *comment* said `scenario.districts`, so the helper's collapse is safe.

**Per-path order preserved.** Parse path: Inv12→Inv4→Inv11→Inv8/9→(Inv1/Inv6 interleaved per-group, then Inv1 events/criteria)→Inv5→Inv2→Inv3→Inv7→terrain — matches original `validateStructural`. Complete path: Inv12→Inv10→Inv11→Inv5→Inv6(full)→Inv1(full)→Inv2→Inv3→Inv4→Inv8/9→Inv7→terrain — matches original `validateScenarioInvariants`. Guards (`if (x !== undefined)`) preserved on the parse side; complete side calls unconditionally as before. Inv10 correctly stays inline (complete-only). Terrain/river block is verbatim (BFS bounds `-2/+2`, clamp `±1`, HEX_DIRS adjacency).

**The two flagged reconciliations are non-observable in production — confirmed at the linchpin.** `parseScenario` calls `validateStructural` (loader.ts:1423), and both production callers run parse first: `loadScenario = validateScenarioComplete(parseScenario(json))` and `generate-scenario.ts` does `parseScenario(partial)` then `validateScenarioComplete(parsed)`. So the parse path always pre-empts, and the complete-path-only changes (Inv7 two-pass→per-precinct; Inv6-before-Inv1) can only surface if a caller invokes `validateScenarioComplete` *without* `parseScenario` first — which nothing in production does.

**Two concerns the "behavior-preserving / agreement" framing understates:**
1. Parse and complete are *not* behaviorally equivalent on distinguishing inputs (e.g. a single group with both an unknown party and a bad vote-share sum: parse throws Inv1, complete throws Inv6). The refactor faithfully preserves each path's own behavior but does not unify them; the test name "agreement" implies an equivalence that doesn't hold on such inputs. Worth a one-line note in the test header.
2. The new agreement tests assert `parse ≡ complete NOW`, so a future edit to a shared helper moves both paths together and the assertion still passes — they cannot catch a both-paths drift from historical behavior. The real historical-equivalence guarantee comes from the pre-existing `loadScenario`-based tests in `loader_test.ts` pinning absolute messages (which exercise the parse path only).

**What I could not verify (no test pins it):** complete-path-*direct* first-error messages for multi-violation Inv7 and the single-group Inv1/Inv6 case *have* changed from historical and are pinned by no test. Latent only for a hypothetical future caller of `validateScenarioComplete` that skips `parseScenario`. Not a regression on any current path.

No defects found. Nice work — the extraction is faithful and the −166 lines remove a real drift hazard.

cgruber: Thanks — thorough equivalence audit, appreciated. Acknowledging the two framing notes (no code change):

1. Test naming: agreed "agreement" is about parse≡complete *now*, not historical equivalence. Historical equivalence is what the pre-existing `loadScenario`-based tests guard; the new block's value is locking the two paths together going forward (and catching a future single-copy edit). I'll leave the name but the distinction is noted.

2. The complete-path-direct first-error for multi-violation Inv7 / single-group Inv1+Inv6 is the one behavior that shifted, and it's unobservable via `loadScenario`/`generate-scenario` (both run parse first) — exactly as documented in the resolved ticket. No current caller hits it; not pinning it with a test is intentional (pinning a non-canonical, non-production message would just re-duplicate logic).

Resolving as LGTM.

PR #302, opened 2026-06-28
fix(game): single-source domain constants + reconcile tie-break (GAME-104)

cgruber:
## Summary

Fixes the **user-visible tie-break contradiction** and single-sources the party set, party
colors, and district palette that had drifted across the codebase (review Theme 5). Resolves
**GAME-104**.

**Tie-break (the correctness bug):** the simulation favored **R** on an exact R/D tie
(`pluralityWinner`, strict `>`) while the adapter's displayed winner favored **D**
(`D >= R ? "D" : "R"`) — a direct contradiction, cemented by two opposing tests. Canonical rule
now: ties resolve **first-in-`ALL_PARTIES`-order (R > D > L > G > I)**, matching the authoritative
simulation. One `winnerOf(share)` helper in `model/types.ts` is used by `election.ts`,
`adapter.ts`, `mapRenderer.ts` (was an *inverted* `['D','R',…]` reduce), and `generator.ts`. Only
visible effect: an exact R/D tie now displays **R** instead of D; district outcomes are unchanged
(`winnerOf` is logically identical to `pluralityWinner`). No shipped scenario has a 50/50
district, so e2e is unaffected.

**Single-sourcing:**
- `ALL_PARTIES` exported from `types.ts`; `main.ts`'s `SPIKE_PARTY_KEYS` literal now imports it.
  The only remaining `["R","D",…]` literal in the codebase is the `ALL_PARTIES` definition.
- `PARTY_COLORS` no longer re-hardcoded in `styles.css`: `main.ts` sets `--party-d`/`--party-r`
  from `PARTY_COLORS` on `:root`; the `.vote-bar` gradient uses `var(--party-d/-r, <fallback>)`.
- `districtColor(id)` in `types.ts` is the single palette lookup (one `#888` fallback,
  deliberately ≠ the unassigned-precinct fill `#2a2a3e`) used by `mapRenderer` + `panels`
  (replaces the `#2a2a3e`/`#888` divergence). `MAX_DISTRICTS = DISTRICT_COLORS.length` (5);
  loader rejects `districts.length > 5` via a single `checkDistrictCount` helper (GAME-102 style).

- The static web game (`game/web`). Behavior change is limited to an exact R/D tie now displaying R, plus a new loud rejection of >5-district scenarios (none ship).

## Validation performed

- [x] `bazel test //game/...` green — **41/41** (incl. e2e). Pre-checked all shipped scenarios: max 5 districts, no 50/50 district — so neither the tie-break flip nor the MAX_DISTRICTS guard changes any shipped behavior.
- [x] Tests added: `winnerOf` (tie→R, highest-share, L/G/I ordering); adapter tie flipped D→R with a comment justifying it by the canonical rule; loader `MAX_DISTRICTS` rejection + inclusive boundary.
- [x] PARTY_COLORS CSS-var fallbacks are byte-identical to the old hardcodes → zero visual change.

## Deployment steps

## Tickets addressed

- Resolves GAME-104 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Reverting restores the prior (contradictory) tie-break and the duplicated constants; no data/schema impact.


game/web/src/model/types.ts
```diff
@@ -6,6 +6,35 @@
 /** Party keys used throughout the sim */
 export type PartyKey = "R" | "D" | "L" | "G" | "I";
 
+/**
+ * Canonical party set and iteration order. The single source of truth for every
+ * winner/margin computation, party-color/label lookup, and party→key mapping.
+ * Iterating this (rather than an inline literal) guarantees no party is silently
+ * dropped and that tie-breaks are deterministic everywhere.
+ */
+export const ALL_PARTIES: readonly PartyKey[] = ["R", "D", "L", "G", "I"];
```
cgruber: **Docstring overclaims a safety property the type doesn't provide; ticket acceptance criterion unmet.**

The docstring says iterating `ALL_PARTIES` "guarantees no party is silently dropped." It does not. `ALL_PARTIES` is typed `readonly PartyKey[]`, which accepts *any subset* of `PartyKey` — `["R", "D"]` (or even `[]`) type-checks here. That is exactly the silent-drop bug class GAME-104 was filed to kill: a literal that omits a party still compiles, and `winnerOf` would then never consider the dropped party.

The ticket's acceptance criterion was the inverse wiring: *"Export `ALL_PARTIES` and derive `PartyKey = typeof ALL_PARTIES[number]`"* — i.e. the array is the source of truth and the union is derived from it, so the union and the array cannot drift. This PR inverted that: `PartyKey` stays a hand-written union and `ALL_PARTIES` is an unchecked, separately-maintained copy. The drop-guard the ticket wanted is absent.

Fix (either):
- Make the array the source: `export const ALL_PARTIES = ["R","D","L","G","I"] as const;` then `export type PartyKey = typeof ALL_PARTIES[number];` (matches the ticket; the array is now exhaustive-by-construction), **or**
- If you intend to keep `PartyKey` hand-written, type the array as a tuple covering every key (e.g. `satisfies readonly [PartyKey, ...PartyKey[]]` won't enforce exhaustiveness either — you need a record-keyed-by-PartyKey pattern). At minimum, drop the "guarantees no party is silently dropped" sentence, because the current typing does not back it.

game/web/src/model/generator.ts
```diff
@@ -152,10 +153,11 @@ export function generatePrecincts(seed = 42): Precinct[] {
...
 
-		// Prior result: winner is plurality party + small margin noise
-		const parties: Array<keyof PartyShare> = ["D", "R", "L", "G", "I"];
-		const winner = parties.reduce((a, b) => (partyShare[a] > partyShare[b] ? a : b));
-		const second = parties
+		// Prior result: winner is plurality party + small margin noise.
+		// Uses the canonical winnerOf / ALL_PARTIES (GAME-104) so the tie-break and
+		// party set match the rest of the codebase.
+		const winner = winnerOf(partyShare);
+		const second = ALL_PARTIES
```
cgruber, line 160: **The generator's tie-break direction changed for non-R/D ties — author's "only an exact R/D tie display changes" is literally false here (low severity, effectively unreachable).**

Old generator winner: `["D","R","L","G","I"].reduce((a,b) => partyShare[a] > partyShare[b] ? a : b)`. A `reduce` with strict `>` keeps `a` only when it is *strictly* greater, so on a tie it advances to the *later* element. Trace a D/L tie (D == L, both highest): old picks **L** (L comes after D in that array). New `winnerOf` scans `ALL_PARTIES = [R,D,L,G,I]` with the same strict `>`, so it keeps the *earlier* element → **D**.

So this is not merely the documented R/D flip; the generator's tie order reversed for any non-R/D tie. Reachability is essentially nil — `generatePrecincts` uses a seeded float RNG and isn't the path for shipped scenarios — so this is a note, not a blocker. But it contradicts the PR's "only an exact R/D tie display changes" framing; the change is correct (it unifies the rule), the claim is just overstated. Worth one line in the PR description so the next reader isn't surprised.

(The `second`-place computation switched array literal to `ALL_PARTIES` too; same earlier-vs-later tie semantics apply to second place, again unreachable in practice.)

game/web/src/model/adapter.ts
```diff
@@ -123,7 +124,9 @@ export function scenarioToSpike(scenario: Scenario): {
 			I: 0,
 		};
 
-		const winner: "D" | "R" = partyShare.D >= partyShare.R ? "D" : "R";
+		// Canonical tie-break (GAME-104): winnerOf scans ALL_PARTIES with strict >,
+		// so an exact R/D tie resolves to R — matching the election simulation.
+		const winner = winnerOf(partyShare);
 		const margin = Math.round(Math.abs(partyShare.D - partyShare.R) * 100) / 100;
```
cgruber, line 134: **Latent inconsistency (low severity, 2-party-safe today): `winner` now ranges over 5 parties but `margin` is hardcoded `|D − R|`.**

`winnerOf(partyShare)` can return any of R/D/L/G/I, but `margin` on the next line is always `|D − R|`. For every shipped scenario this is fine — the adapter only ever populates R and D (lines 119–125 hardcode `L/G/I = 0`, and it maps only `parties[0]→R`, `parties[1]→D`). So the day a scenario with a real third party ships, `winner` could be L while `margin` still reports the D−R gap — a wrong number, not just a wrong color.

Not blocking for this PR (no such scenario exists, and adapter L/G/I are structurally 0). Flagging because the PR's "only an exact R/D tie display changes" claim glosses over this: the winner computation widened to 5 parties without the margin following. Either add a comment that this adapter is 2-party-only by construction, or compute margin against the actual second-place share (as election.ts does) so the two stay consistent.

game/web/src/model/types.ts
cgruber: Fixed properly. Made the array the source of truth and DERIVED the type: `ALL_PARTIES = ["R","D","L","G","I"] as const` and `PartyKey = (typeof ALL_PARTIES)[number]`. Now omitting a party from the array narrows `PartyKey`, so every `Record<PartyKey, …>` (PARTY_COLORS, PARTY_LABELS) and PartyShare-keyed access becomes a compile error — the no-silent-drop guarantee is real, matching GAME-104's intent. Updated the docstring to describe the actual mechanism. bazel test //game/... still 41/41 (the `as const` readonly tuple didn't ripple — `.slice().sort()` in election.ts still yields a mutable array, all iteration is read-only).

game/web/src/model/generator.ts
cgruber: Acknowledged — fair catch on the claim wording. `generator.ts` is dead code (no importers; slated for delete/deprecate in GAME-109), so this tie-direction change is not runtime-observable. I kept the `winnerOf` call here rather than reverting, because leaving the inverted `["D","R",…]` literal in place would contradict GAME-104's single-source goal and would just have to be re-touched when generator is removed. To be precise: the only RUNTIME-observable behavior change in this PR is the exact-R/D-tie display flipping D→R; this generator (dead-path) non-R/D tie normalization is not reachable. I'll leave the PR summary's "district outcomes unchanged" (true) and note this dead-path nuance here for the record.

game/web/src/model/adapter.ts
cgruber: Added a comment. This adapter is the 2-party spike path — L/G/I are hardcoded to 0 and only the first two scenario parties are mapped, so `|D − R|` is the exact winner-vs-runner-up margin today. Documented that a future 3+-party adapter must compute the margin against actual second place. Not changing the computation now (no 3-party scenario exists; that's GAME-043's spike-layer retirement territory).

PR #305, opened 2026-06-28
fix(a11y): focus management, dialog/modal semantics, keyboard-safe tutorial lock (GAME-105)

cgruber:
## Summary

Fixes the keyboard / screen-reader navigation model (review Theme 4). Before this: `.focus()`
was never called anywhere (WCAG 2.4.3), overlays weren't dialogs and weren't focus-trapped (a
keyboard user could Tab into the **live editor behind the result modal** and paint / re-submit),
and the tutorial lock was pointer-events-only (keyboard bypassed the freeze). Resolves
**GAME-105**.

- **Focus management:** `focusEl`/`focusFirst`/`setEditorInert` helpers (main.ts) wired into every
  transition — menu→first enabled nav button, editor→#map-svg, result→Keep Drawing,
  wrap-up/debrief→headings (`tabindex="-1"` where needed), about, campaign/scenario select, WIP
  modal. Applied via rAF after un-hide.
- **Dialog semantics + inert:** `#about-screen`/`#wip-warning-modal`/`#result-screen`/
  `#wrap-up-screen` get `role="dialog" aria-modal="true"` + label. Result/wrap-up set
  `#app-header`+`#main` `inert` (cleared on Keep Drawing — the only in-page exit; every other exit
  reloads), so keyboard focus can't fall through to the live editor behind the modal. WIP modal
  inerts `#scenario-select` (cleared on Cancel).
- **Keyboard-safe tutorial lock:** `setMapKeyboardLock` (overlay.ts) sets `inert`+`tabindex=-1`+
  blur on `#map-svg` for non-paint steps; paint steps keep map+toolbar live. Defense-in-depth: the
  map keydown paint handler bails on `#map-svg[inert]` and the document-level zoom shortcut bails
  on `closest('[inert]')` — `inert` is an HTMLElement property and isn't enforced on SVG roots by
  Chromium. (The zoom shortcut was a real leak — bound to `document`, it fired behind modals /
  during frozen steps.)
- `#nav-back-menu`: first menuitem focused on open, ArrowUp/Down roving, Escape returns focus.
- `renderScenarioCards` sets `role="listitem"`; key-range aria-labels built from `districtCount`
  (was hardcoded "1–5"; e.g. scenario-002's 4 districts → "1–4").

- The static web game (`game/web`). Presentation/a11y only — no game logic changed. User-perceptible behavior change: focus now moves on transitions, and the editor is inert behind the result/wrap-up modals.

## Validation performed

- [x] `bazel test //game/...` green — **42/42** (e2e: **170 Playwright tests**, including 7 new GAME-105 tests).
- [x] New a11y tests do NOT pre-focus (the flaw in the old suite): submit→`activeElement` inside `#result-screen`; editor `inert` behind result + `#map-svg` unfocusable; Keep Drawing clears inert; tutorial non-paint step blocks keyboard paint (assignment count unchanged); `role="listitem"`; dynamic key-range label.
- [x] Mentally + e2e-verified the inert round-trip: menu→intro→editor→submit→result→Keep Drawing→editor (inert cleared, editor usable again).
- [ ] N/A — no infra/deploy change.

## Deployment steps

## Tickets addressed

- Resolves GAME-105 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Restores the prior (no-focus-management, non-modal, keyboard-bypassable) behavior; no data/schema impact.


game/web/e2e/a11y.spec.ts
```diff
@@ -143,6 +143,184 @@ test.describe("GAME-008 Accessibility", () => {
...
+      return store ? store.getState().assignments.size : -1;
+    });
+
+    // Attempt to focus the map and drive the number-key paint handler.
+    await page.evaluate(() => document.getElementById("map-svg")?.focus());
+    await page.keyboard.press("ArrowDown");
+    await page.keyboard.press("2");
+    await page.waitForTimeout(150);
+
+    const after = await page.evaluate(() => {
```
cgruber: **The keyboard-lock test is vacuous — it would NOT fail if the lock were removed.** (high)

This test asserts `expect(after).toBe(before)` where `before`/`after` are `store.getState().assignments.size`. But the adapter seeds an entry for **every** precinct at load time — `adapter.ts:169-173` does `assignments.set(i, spikeDistId)` for all precincts (null for unassigned). And `paintStroke` (`gameStore.ts:90-100`) only mutates *values* on already-present keys (`next.set(id, district)`), never adding keys. So `assignments.size` equals the precinct count from load and is **invariant under painting**.

Consequence: if the keyboard lock (`setMapKeyboardLock`) were deleted, `ArrowDown`+`2` would successfully paint precinct 0 into district 2 — yet `assignments.size` would still be unchanged, so `after === before` still holds and the test still **passes**. The test does not discriminate the bug it claims to be the "PRIMARY assertion (the literal ticket requirement)." This is the exact gap the old pre-focusing suite had, reproduced in a new form.

Fix: assert on assignment *values*, not `.size`. E.g. capture the focused precinct's district before/after (`store.getState().assignments.get(focusedId)`), or count non-null entries: `[...assignments.values()].filter(v => v != null).length`. Pick a measure that changes when a precinct is painted. To be sure it discriminates, confirm the same `ArrowDown`+`2` sequence *does* change that measure on an unlocked map (e.g. a paint step or a non-tutorial editor).

cgruber: Follow-up on the fix: two of the alternatives I listed can themselves be vacuous, so prefer the snapshot form:

- `get(focusedId)` / `count non-null` both fail if precinct 0 starts non-null, and `paintStroke` no-ops entirely when precinct 0's initial district already equals 2 (`gameStore.ts:96` `next.get(id) !== district` guard).
- Note the painted precinct is **precinct 0**, not a neighbor: the first `ArrowDown` is consumed by the `focusedPrecinctId === null` init at `mapRenderer.ts:1318`, so the subsequent `"2"` paints `precincts[0]`.

Foolproof: snapshot all entries and compare —
`JSON.stringify([...store.getState().assignments.entries()])` before vs after, assert equal — paired with a control proving the same `ArrowDown`+`2` sequence *does* change that snapshot on an unlocked map (pick a precinct/district pair whose initial value differs from the painted one).

cgruber: Fixed exactly as recommended. The locked test now fingerprints the assignment VALUES (`JSON.stringify([...assignments.entries()])`), not `.size` — so a real paint would change it. And I added a companion **control** test ("control: keyboard number key paints on an unlocked map", scenario-002, no tutorial → map not inert) that runs the identical sequence (focus → ArrowDown inits focus to precinct 0 → number key) and asserts the fingerprint DID change. Together they prove the lock blocks a real paint, not a no-op sequence. (Verified the control's `1` actually paints precinct 0: paintStroke ignores `editable` and precinct 0 is null-seeded, so null→1 flips the fingerprint.) e2e green.

cgruber: Agreed — I used the snapshot-of-all-entries form precisely to avoid the count-non-null / get(0) vacuity traps you flagged. `JSON.stringify([...entries()])` captures any value flip on any key, and the control test guarantees the sequence is a genuine paint, so neither side can pass vacuously. Thanks for the catch.

PR #306, opened 2026-06-28
test(game): test-infra backstops + missing coverage (GAME-109)

cgruber:
## Summary

Adds structural test backstops so coverage can't silently erode, plus the missing unit/e2e tests
(review Theme 6). Mostly additive. Resolves **GAME-109**.

- **Meta-test `//game/web:unwired_tests_test`** (sh_test, `game/web/unwired_tests_test.sh`): fails
  when a `src/**/*_test.ts` exists with no `js_test` wiring it — the silent-coverage-gap the
  colocated-test convention invites. Bazel `glob` doesn't cross packages, so each subpackage
  exposes a `:test_sources_for_meta` filegroup; the script matches each test's `<name>_test.js"`
  entry_point in the BUILD blob, guards the vacuous zero-file case, and is bash-3.2 portable.
  **Dummy-proved** (a temporary unwired `_test.ts` makes it fail naming that file). Backstop only —
  keeps the explicit per-target convention. (Documented caveat: a future NEW subpackage must
  register its own filegroup.)
- **`loader_integration_test.ts`** now globs the scenarios dir from `RUNFILES_DIR` (excludes
  `*.overlay.json`) instead of a hardcoded 12-entry table; known counts kept via an `EXPECTED`
  lookup. Picks up every shipped scenario (incl. `tutorial-001`, which the old table omitted) — a
  new scenario can no longer escape content validation.
- **`evaluate_test.ts`**: `safe_seats` (margin above / `==` / below threshold; `safeCount < min_count`)
  and `competitive_seats` boundary cases.
- **New `criterion_icons_test`** (+ targets): icon per `Criterion` type, `validity:*` prefix
  resolution, and a compile-time `Record<Criterion["type"], true>` exhaustiveness guard (adding a
  type without an icon fails to compile). `criterion-icons.ts` excluded from the app glob to avoid
  an output collision, re-added via deps.
- **Deleted dead `generator.ts` + `generator_lib`** (verified no importers; `generatePrecincts`
  unused) and dropped the vestigial `generator_lib` dep from `adapter_test` + app typecheck/esbuild.
- **Result-screen verdict DOM e2e** (raised in the GAME-110 review): a winning scenario-002
  submission asserts `#result-stars` visible with filled stars + verdict/subtitle text; a failing
  submission asserts `#result-stars` hidden + "Map Failed".

- The static web game (`game/web`) build/test graph. New test targets + one deleted dead source; no runtime behavior change.

## Validation performed

- [x] `bazel test //game/...` green — **44/44** (the new meta-test, criterion_icons_test, the globbed integration test, and the e2e additions all pass).
- [x] Meta-test dummy-proof: a temporary `src/model/zzz_unwired_test.ts` made it FAIL naming that file; removed → passes.
- [x] `generator.ts` confirmed unimported before deletion (no `import` of `./generator`/`model/generator`, `generatePrecincts` unused); `//game/web:app` + `:deployable` still build.

## Deployment steps

- N/A — tests + dead-code removal; no runtime/deploy change.

## Tickets addressed

- Resolves GAME-109 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Removes the backstops + added tests and restores `generator.ts`; no data/schema impact.


game/web/unwired_tests_test.sh
```diff
@@ -0,0 +1,107 @@
...
+while IFS= read -r tf; do
+  [ -n "$tf" ] || continue
+  base="$(basename "$tf")"          # e.g. loader_integration_test.ts
+  entry="${base%.ts}.js"            # e.g. loader_integration_test.js
+  # A wired test has its js entry_point quoted in some BUILD file. Match the
+  # basename followed by the closing quote, so a directory-prefixed entry_point
+  # like "src/criterion-icons_test.js" still matches, while loader_test.js does
+  # NOT satisfy loader_integration_test.js (the leading char before the basename
+  # is a quote or a path separator, never a name char).
+  if printf '%s' "$BUILD_BLOB" | grep -qF "$entry\""; then
```
cgruber: **Meta-test false-match: the wired-check anchors only the *right* side, so a shorter entry that is a suffix-substring of a longer wired entry is falsely reported as wired.**

`grep -qF "$entry\""` matches `<basename>.js"` anywhere in the BUILD blob, anchoring only the trailing quote — never the character *before* the basename. So an unwired `foo_test.ts` (entry `foo_test.js"`) is reported "wired" whenever some *other*, unrelated wired test ends in `…foo_test.js"`.

This is not hypothetical — there's a live collision today:

```
$ grep -c 'stage_test.js"' game/web/src/pipeline/BUILD.bazel
3   # population-stage_test.js", county-stage_test.js", demographics-stage_test.js"
```

A hypothetical unwired `src/pipeline/stage_test.ts` (entry `stage_test.js"`) would match those three wired entries and pass — the exact silent-coverage-gap this backstop exists to catch.

The block comment above this line (lines 95-99, "...so a directory-prefixed entry_point like `src/criterion-icons_test.js` still matches, while `loader_test.js` does NOT satisfy `loader_integration_test.js` (the leading char before the basename is a quote or a path separator, never a name char)") asserts an invariant the code does not enforce: `grep -qF "$entry\""` does not check the leading char at all. The `loader_test`/`loader_integration_test` pair happens not to collide (there's `_integration_` between the prefix and `_test.js`), but `-stage_test.js"` proves the general claim false — the char before the match is `-`, a name char.

Fix: anchor the left side to the two real entry_point forms (bare basename or `<dir>/<basename>`) — quote-or-slash before, quote after:

```sh
esc=$(printf '%s' "$entry" | sed 's/[.[\*^$/]/\\&/g')
if printf '%s' "$BUILD_BLOB" | grep -qE "[\"/]${esc}\""; then
```

Severity: medium / latent. No current test is unwired and no current file triggers the false pass, so nothing is broken today — but it directly weakens the single guarantee the meta-test is meant to provide, and the next two-word test name that suffixes an existing one (e.g. a future `stage_test.ts`, `store_test.ts`) silently re-opens the gap. Worth fixing while the mechanism is fresh.

cgruber: Fixed — strong catch. The wired-check now anchors BOTH sides: `grep -qE "[\"/]${entry_re}\""` (left = `"` for a bare entry_point or `/` for a directory-prefixed one like `"src/criterion-icons_test.js"`; right = closing `"`; `.` escaped). Your exact collision is now caught: an unwired `stage_test.ts` no longer matches `"population-stage_test.js"` because the char before the basename there is `-`, not `"`/`/`. Re-ran the meta-test: all 24 wired tests still detected (no false-miss, incl. the `/`-prefixed criterion-icons case). Updated the misleading comment to describe the actual two-sided anchoring.

PR #307, opened 2026-06-28
fix(deploy): fail loud on broken index.html patch + durable per-env idempotency (GAME-108)

cgruber:
## Summary

Hardens the deploy tool (`release.main.kts`) and the deployable genrule (review Theme 7 / Tooling).
The deploy tool is **not** exercised by `bazel test`, so the changes are conservative and
reviewable by reading; a new CI anchor-test guards the patch literals. Resolves **GAME-108**.

- **Hard-fail on failed index.html patches:** the three exact-string patches (bundle cache-bust
  `?v=$version`, app-version meta, app-environment meta) now `err()`→`exitProcess(1)` **before any
  commit/push**, instead of warning and continuing (which shipped a stale bundle / empty
  version+env meta and exited 0). A missing index.html also errors. After writing, a post-write
  readback asserts every patched string is present.
- **Durable per-env idempotency:** the already-deployed guard now reads the committed per-env
  `deployment-metadata.json` (`jj file show -r web_deploy <env>/deployment-metadata.json`, root for
  production) instead of scraping the tip commit description. The tip-scrape was wrong — each
  deploy commits only its own env line, so staging→beta→staging found no staging match on the tip
  and re-deployed. Nonzero/blank/unparseable → proceed (first deploy to a fresh env never blocked).
- **Genrule asset guard:** wrapped the basename `case` with `case "$f" in */assets/*) ;; *)
  <existing> ;; esac` so a future `assets/*.json` (sprite atlas/config) or `*.wasm` isn't misfiled
  into `scenarios/` or root. Zip contents byte-identical today (no such asset exists yet).
- **CI anchor test:** new `//game/web:release_patch_anchors_test` (sh_test) asserts index.html
  still contains the exact literals `release.main.kts` greps for, so an index.html reformat fails
  CI rather than production.

- The deploy tooling (`game/release.main.kts`) + the deployable zip layout. No game runtime change. The deploy behavior change is fail-loud on a broken patch + a correct (not over-eager) idempotency skip.

## Validation performed

- [x] `bazel test //game/...` green — **45/45** (incl. the new `release_patch_anchors_test` + e2e).
- [x] `bazel build //game/web:deployable` clean; zip contents verified unchanged (index.html/styles.css/bundle.js/react-stub.js/wasm/scenarios + assets tree).
- [x] Anchor test proven to FAIL when an anchor is mangled (tested against a scratch copy; real index.html untouched).
- [x] `release.main.kts` changes reviewed by reading (the tool is not bazel-test-exercised and must not be run in a test): patch hard-errors + post-write assertion + per-env metadata guard all verified against where the deploy writes `deployment-metadata.json`.
- [ ] N/A — could not run an actual deploy (would mutate `web_deploy`); not run.

## Deployment steps

- This changes the deploy tool itself. The next `game/release.main.kts -- deploy` run picks it up; behavior is stricter (aborts on a broken index.html patch) — no migration. The per-env guard reads `deployment-metadata.json` which the deploy already writes, so existing envs work on next deploy.

## Tickets addressed

- Resolves GAME-108 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Restores the warn-and-continue patch behavior + tip-commit idempotency scrape + the unguarded genrule; no data/schema impact.


game/release.main.kts
```diff
@@ -324,11 +324,29 @@ class Deploy : CliktCommand(
...
+        //
+        // Assumes `jj file show` (modern jj; was `jj cat`) — this repo is modern jj. A nonzero
+        // exit / blank / unparseable result is treated as "not yet deployed → proceed" so the
+        // FIRST deploy to an env (no file yet) is never blocked.
         if (!isTestBuild) {
-            val (log) = trysh("jj", "log", "-r", "web_deploy", "--no-graph", "-T", "description")
-            val deployed = Regex("$env: ([^ ]+)").find(log)?.groupValues?.get(1) ?: ""
-            if (deployed == version) {
+            val metaPath = if (env == "production") "deployment-metadata.json" else "$env/deployment-metadata.json"
+            val (body, exit) = trysh("jj", "file", "show", "-r", "web_deploy", metaPath)
```
cgruber: **HIGH — idempotency guard reads a path that never exists; the fix is silently defeated.**

`trysh(...)` here passes no `dir=`, so it defaults to `dir = gameDir` (see `trysh` signature, ~line 97), i.e. `jj` runs from `game/`. But `jj file show` resolves fileset path arguments **relative to cwd**, not the repo/web_deploy tree root. The metadata files live at the *top* of the `web_deploy` tree (`deployment-metadata.json`, `<env>/deployment-metadata.json`), which corresponds to `repoRoot` — not `game/`.

So from `game/` the read becomes `game/deployment-metadata.json` (or `game/beta/deployment-metadata.json`), which doesn't exist:

```
$ cd game && jj file show -r web_deploy deployment-metadata.json
Error: No such path: deployment-metadata.json   (exit 1)
$ cd game && jj file show -r web_deploy beta/deployment-metadata.json
Error: No such path: beta/deployment-metadata.json   (exit 1)
```

`exit != 0` ⇒ `deployedVersion = ""` ⇒ `"" == version` is always false ⇒ the skip branch is **never** taken. The guard fail-opens on *every* deploy — a re-deploy of the same version is no longer caught. This is precisely the "reading a path that never exists → every deploy re-deploys, silently defeating the fix" mode. (The old `jj log -r web_deploy -T description` took no path arg, so cwd was irrelevant; this cwd sensitivity is new to this PR.)

Verified the same command from `repoRoot` returns the live production metadata at exit 0 — so the tree root is `repoRoot`.

**Fix:** pass `dir = repoRoot` to this `trysh` call:
```kotlin
val (body, exit) = trysh("jj", "file", "show", "-r", "web_deploy", metaPath, dir = repoRoot)
```
(`repoRoot` is a top-level val, in scope here.) Worth a quick add to the comment that the path is web_deploy-root-relative so this doesn't regress.

cgruber: Fixed — excellent empirical catch. Added `dir = repoRoot` to the `jj file show` call. You're exactly right: `trysh` defaults to `dir = gameDir` (game/), and `jj file show` resolves the fileset path relative to cwd, so from game/ the web_deploy-root-relative path (`beta/deployment-metadata.json`, root `deployment-metadata.json`) never resolved → exit 1 → fail-open → the guard never skipped. Verified the fix the same way you found the bug: `jj file show -r web_deploy deployment-metadata.json` from repoRoot returns the live production metadata (`version: coming-soon`, `environment: production`), so the guard now reads the real deployed version. Added a comment explaining the cwd dependency so it can't regress. (`dir` is an existing named param on `trysh`, so it compiles.)

PR #308, opened 2026-06-29
build(game): add Biome lint + format CI gate (GAME-107)

cgruber:
## Summary

Adds the first **linter + formatter + CI gate** to `game/web` — the production game had none (only
the strict tsconfig; review Theme 7 meta / top-10 #10). This is the systemic backstop for the
non-null-assertion / literal-drift class and turns the GAME-108 index.html reformat hazard into a
CI failure rather than a production one. Resolves **GAME-107**.

- **Biome 1.9.4** (matches the spike's `biome.json`) added to `game/package.json`; `setup.sh`'s
  `npm ci` installs it (same path as the playwright e2e binary).
- **`game/web/biome.jsonc`:** formatter tuned to the **existing** house style (tab indent, double
  quotes, semicolons, `trailingCommas: all`, `lineWidth 100` — empirically the smaller diff vs 120)
  so the mechanical reformat is minimal. `recommended` lint rules enforced as **errors**;
  high-volume *intentional-pattern* rules deferred (`"off"` + a `TODO GAME-107` with count +
  rationale): `noNonNullAssertion` (~401, the `spec.count!` optional-field pattern), `useLiteralKeys`
  (~214 — its autofix `obj["k"]→obj.k` breaks `noUncheckedIndexedAccess`/`exactOptionalPropertyTypes`),
  `noExplicitAny` (4), and a few unsafe-autofix rules (`noForEach`, `noDelete`, `useTemplate`, …).
  Green now, tightenable incrementally rather than via a big refactor.
- **Reformat:** `biome check --write` touched **73 files** (the tree was mixed tab / 2-space). Purely
  mechanical EXCEPT one hand-fixed formatter bug: Biome wrapped inline `import("./scenario.js").CompareOp`
  type expressions into invalid TS (TS1005) in `loader.ts` → refactored to a named `CompareOp` import
  (behavior-identical cast).
- **Gate:** `//game/web:lint_test` (sh_test mirroring `e2e_test` — `local=True`, `no-sandbox`, runs
  the physical `node_modules/.bin/biome ci .`). Per-subpackage `:lint_sources` filegroups make every
  linted file a data dep (cache-sound; `glob` doesn't cross packages). **Proven to FAIL** on a
  violation twice (a subpackage file + an e2e file), then reverted.
- **Docs:** new `game/web/README.md` (house style + biome commands + the gate + deferred-rules note);
  `setup.sh` final message points at the lint/fix commands.

- `game/web` build/test graph + dev tooling. The 73-file reformat is mechanical (whitespace/quotes); no runtime behavior change. One semantic-adjacent fix (`CompareOp` import) is a behavior-identical type cast.

## Validation performed

- [x] `bazel test //game/...` green — **46/46** (45 prior — typecheck, unit, e2e — plus the new `lint_test`). The reformat broke no test.
- [x] `biome ci .` exits 0 on the clean tree; the gate **fails** on an introduced violation (verified twice, then reverted).
- [x] The one non-mechanical change (`CompareOp` named import in `loader.ts`) typechecks — same cast, 4 sites.

## Deployment steps

- N/A — tooling + mechanical reformat; no runtime/deploy change. (Going forward, `bazel test //game/web:lint_test` gates style; devs run `biome check --write .` locally — see `game/web/README.md`.)

## Tickets addressed

- Resolves GAME-107 (file removed, moved to Resolved in `TICKETS.md`).

## Rollback

- Revert this commit. Removes the gate + config and restores the prior (mixed) formatting; no data/schema impact.


game/web/BUILD.bazel
```diff
@@ -219,6 +219,61 @@ sh_test(
...
+        "tsconfig.json",
+        "tsconfig.lib.json",
+        "playwright.config.ts",
+        "//game/web/src/model:lint_sources",
+        "//game/web/src/simulation:lint_sources",
+        "//game/web/src/pipeline:lint_sources",
+        "//game/web/src/store:lint_sources",
+        "//game/web/src/audio:lint_sources",
+        "//game/web/src/testing:lint_sources",
+    ] + glob([
```
cgruber, line 259: **Gate erosion gap: a future subpackage's `:lint_sources` is not forced into `lint_test` `data` (medium).**

All 6 current subpackages (model, simulation, pipeline, store, audio, testing) are correctly wired into the `data` list, and I verified each `:lint_sources` filegroup globs `*.ts` over its whole package. No current coverage hole — confirmed by a prove-it-fails: editing a file in `src/store/` (reached only via `//game/web/src/store:lint_sources`) invalidated the cache and failed the gate *without* `--nocache_test_results`, so the per-subpackage data-dep wiring genuinely invalidates on edit.

The gap is for *new* packages. `biome ci .` lints the physical tree, so a new `src/newpkg/` with its own `BUILD.bazel` still gets linted on a cold run — but because its `:lint_sources` is not in this `data` list, **edits to it won't invalidate the `lint_test` cache** → a violation introduced there serves a stale PASS until an unrelated cache bust. This is exactly the erosion class GAME-109's `unwired_tests_test` guards for *test* files, but nothing guards the lint `data` list against a forgotten new subpackage.

Suggested fix (non-blocking): a small meta-test (mirroring `unwired_tests_test`) that fails when a `game/web/src/**/BUILD.bazel` package's sources are absent from `lint_test`'s `data`, OR a comment in each new-package BUILD template reminding to add `:lint_sources`. Filing as a follow-up ticket is fine; this does not block the PR.

cgruber: Good catch, and agreed it's non-blocking — all 6 current packages are wired (verified) and this only bites a *future* new `src/<pkg>/` subpackage, which would lint-clean on a cold run but not invalidate on edit until an unrelated cache bust. It's the same gate-erosion class GAME-109's `unwired_tests_test` guards for `*_test.ts`, so the right fix is a sibling meta-test asserting every `game/web/src` subpackage BUILD that has `*.ts` exposes a `:lint_sources` filegroup that's in `lint_test`'s `data`. Tracking it as a follow-up ticket (GAME-111) rather than expanding this already-large reformat PR. Resolving.

PR #318, opened 2026-07-01
refactor(model): unify type systems — party-agnostic runtime model (GAME-043 PR 1 of 2)

cgruber:
## Summary

Core of **GAME-043** (the largest refactor in the backlog): retire the two parallel type systems —
the spike `model/types.ts` (fixed `PartyKey`/`PartyShare {R,D,L,G,I}`) and the `adapter.ts`
translation layer — in favor of **one party-agnostic runtime model keyed by the scenario's
arbitrary `PartyId`**. This is what makes multiparty (GAME-112) fall out. Follows the approved
design (`thoughts/shared/plans/2026-06-30-game-043-unify-types-design.md`). **PR 1 of 2** —
`Party.color` authoring is PR 2, so this does **not** resolve the ticket.

**Behavior-preserving:** all 12 shipped scenarios are two-party, so a model that aggregates all
parties is provably equivalent today; the existing 47 tests are the regression bar.

- **New `model/runtime.ts`** (unified runtime `Precinct` — numeric `index` + string `scenarioId`,
  `voteShare: Record<PartyId,number>`, `previousResult`, terrain/neighbors; dummy `demographics`
  dropped; `DistrictResult`/`SimulationResult`/`GameState` party-id-keyed, `GameState` carries
  `parties`) and **`model/party.ts`** (`winnerOf(share, parties)` with first-in-scenario-order
  tie-break, `zeroShare`, `PARTY_PALETTE` + `partyColor` by party index). **`model/types.ts`
  deleted.**
- **`adapter.ts` → a builder** (`scenarioToRuntime`): aggregates ALL parties (pop-weighted, no
  turnout), margin vs. the actual sorted runner-up, emits `index`/`scenarioId` + `parties`.
- Every hardcoded `.R`/`.D` site migrated; `partyIdToKey` **deleted** (evaluate criteria use
  `PartyId` directly); `efficiency_gap` kept two-party via `parties[0]/[1]` (the proper
  two-party-normalization is GAME-112). Colors via `PARTY_PALETTE` by party order (ken=orange,
  ryu=purple — look preserved).
- BUILD: `types_lib` removed; `runtime_lib` + `party_lib` added; deps/data flipped across
  model/simulation/store + game/web.

- The static web game (`game/web`) — model/simulation/render/store/main + BUILD. Behavior-preserving refactor.

## Validation performed

- [x] `bazel test //game/...` green — **47/47** (all unit + typecheck + winnability e2e).
- [x] `winnerOf` tie-break hand-verified (seeds `parties[0]`, strict `>`, first-in-scenario-order — reproduces the old R>D>L>G>I determinism).
- [x] Test-fixture changes reviewed: `validity_test`'s line delta is the dropped dummy `demographics` field + field renames, not removed assertions.
- [x] Flagged behavior deltas confirmed inert today: `voteShare` no longer carries `0` for parties absent from the scenario (consumers guard `?? 0`; all scenarios 2-party); `PARTY_LABELS` fallback slots 3–5 only reachable by an unnamed >2-party scenario (none exist).

## Deployment steps

- Standard: ships with the next deploy. No special steps (behavior unchanged).

## Tickets addressed

- Advances GAME-043 (PR 1 of 2; `Party.color` authoring is PR 2 — ticket stays open).

## Rollback

- Revert this commit. Restores the spike types / adapter translation layer; no data/schema impact.


game/web/src/simulation/evaluate.ts
```diff
@@ -249,33 +254,37 @@ export function evaluateCriteria(
...
+						p2Wasted += p2Votes;
 					}
 				}
-				const rawGap = allVotes > 0 ? (rWasted - dWasted) / allVotes : 0;
+				const rawGap = allVotes > 0 ? (p1Wasted - p2Wasted) / allVotes : 0;
 				const absGap = Math.abs(rawGap);
 				passed = applyOp(absGap, c.operator, c.threshold);
-				const direction = rawGap >= 0 ? "R-disadvantaged" : "D-disadvantaged";
+				const direction = rawGap >= 0 ? `${party1}-disadvantaged` : `${party2}-disadvantaged`;
 				detail = `efficiency gap: ${(absGap * 100).toFixed(1)}% (${direction}; required ${OP_LABEL[c.operator]}${(c.threshold * 100).toFixed(0)}%)`;
```
cgruber, line 288: **Low severity — untested user-visible output change (not a masked regression, but flagging per behavior-preservation audit).**

The criterion `detail` strings changed keyspace along with the logic: this block now emits the scenario party-id (`${party1}`/`${party2}`, e.g. `ken-disadvantaged`) where the pre-GAME-043 code emitted the mapped spike key (`R-disadvantaged`/`D-disadvantaged`). Same lateral change in `seat_count`, `safe_seats`, and `mean_median` details (`${c.party}` replaces the `partyIdToKey`-mapped `key`).

The pass/fail computation is fully behavior-preserving (verified: `p1Wasted`/`p2Wasted` over `parties[0]`/`parties[1]` == old `rWasted`/`dWasted` over R/D, abs() so direction is irrelevant to the gate). Only the **displayed detail text** differs.

I confirmed **no test pins these strings** — grep of `evaluate_test.ts` for `disadvantaged`/`detail`/`seat(s)` finds only test descriptions and pass/fail assertions, none asserting `detail` content. CI-green therefore can't observe this delta. So it's genuinely unpinned.

Net: this is a pre-existing lateral non-equivalence (both old and new leak an internal token rather than the display name — `R` was never the player-facing label either), acceptable for a structural PR, but it IS a real output-text change on the "behavior-preserving" claim. Worth a one-line note in the PR body / GAME-112 so the multiparty PR doesn't assume these strings were always party-id. No change requested.

game/web/src/model/runtime.ts
```diff
@@ -1,66 +1,25 @@
...
-/** Sex/gender demographic breakdown; floats summing to 1.0 */
-export interface Demographics {
-	male: number;
-	female: number;
-	nonbinary: number;
-}
+// ─── Geometry ─────────────────────────────────────────────────────────────────
 
 /** Axial hex grid coordinates (cube system, q+r+s = 0) */
 export interface HexCoord {
```
cgruber, line 25: ## Independent behavior-preservation audit — GAME-043 PR 1 of 2

**Verdict: LGTM** (one low-severity untested-output note posted inline on evaluate.ts:288; not a change request).

Base compared against: `dc427d1c1680` (PR parent). I audited every migrated site algebraically and every changed test assertion-by-assertion. The "behavior-preserving for shipped (2-party) scenarios" claim holds.

### Keystone verified (the antecedent the whole claim rests on)
All 12 shipped scenarios have **exactly 2 parties** (`jq '.parties|length'`): scenario-002..008 + tutorial-001..004 = `ken`/`ryu`; scenario-009 = `cat`/`dog`. (`scenario-002.overlay.json` has no `parties` — it's an overlay, not a scenario.) So every divergence that only appears at N≥3 (margin-vs-runner-up, efficiency_gap over `parties[0/1]`, lean `p2−p1`) is unreachable today. scenario-009's cat/dog is a nice proof the old R/D-indirection was doing nothing a 2-party scenario needs.

### Migrated sites — verified equivalent for 2-party
- **winnerOf** (party.ts): iterates `parties` with strict `>`, seeds `parties[0]`. Since old `firstParty→R→ALL_PARTIES[0]`, tie-break order is identical by construction. Absent-key `?? 0` is safe.
- **adapter margin**: `voteShare[winner] − voteShare[runnerUp]` (sorted) == old `|D − R|` (winner has max, runnerUp second → non-negative; exact tie → 0 both ways). Same rounding (shares→3dp, margin→2dp).
- **election.ts**: `voteTotals` over `parties`, `runnerUp = sorted[1] ?? parties[0]` == old `?? "R"`. seatsByParty keyed by winner PartyId.
- **evaluate.ts efficiency_gap**: `p1/p2` over `parties[0]/[1]` == old `R/D`; abs() gates so direction irrelevant. seat_count/safe_seats/mean_median now key `c.party` directly on the (now party-id) seatsByParty/voteTotals — consistent keyspace, drops the `partyIdToKey` indirection cleanly.
- **mapRenderer lean**: `p2Share − p1Share` == old `partyShare.D − partyShare.R` (party2↔D, party1↔R). Same sign/value → same PuOr `t`.
- **panels result card**: renders party2 first with `--d-pct:${p2Pct}%` then `p2Label · p1Label` — identical order/values to old `--d-pct:${dPct}%` / `dLabel · rLabel`. winnerColor via `partyColor` == old `PARTY_COLORS` (palette-by-index, party1→orange, party2→purple).
- **main.ts CSS vars**: `--party-r=PALETTE[0]`, `--party-d=PALETTE[1]` == old `PARTY_COLORS.R/.D`.

### Test-change audit — NO asserted value changed
adapter_test / election_test / evaluate_test / validity_test are **keyspace-only rewrites** (`.R`→`voteShare[pid]`, `.id`→`.index`, `"R"`→`pid("r_party")` or `PARTIES[0]`). Every winner, margin (0.4/0.6/0.02/0.2), share (0.7/0.3, 0.56/0.44), and seat count (R=1/D=2), and every tie-break case (5-way→R, D>L, L>G, G>I) is **unchanged**. election_test cleverly reuses the literal `["R","D","L","G","I"]` strings as PartyIds so the fixtures read identically. The only *value* delta is intentional and correct: `p.partyShare.L === 0` → `p.voteShare[l_party] === undefined` (absent-party design). No assertion weakened, removed, or loosened.

### Guards / dead-key
Every production `voteShare`/`voteTotals` access is `?? 0`-guarded (no NaN). `seatsByParty` is only accessed by direct key (`[r.winner]`, `[c.party]`) — never enumerated (`Object.keys/entries/values`/`for-in`), so the "absent parties have no key" change can't cause a missing-iteration bug.

### CAT:PERSISTENCE/CODEC — no back-compat break
- **zundo undo/redo**: `partialize` captures only `{assignments, simulationResult}` — never precincts/voteShare/demographics. Snapshots are in-memory (session lifetime), never serialized. Key renames + dropped demographics + new `parties` field cannot touch undo.
- **WIP save** (progress.ts `WIP_KEY`): persists only `{scenarioId, assignments: Record<string,number> (index-keyed), activeDistrict}`. No precinct shape stored. `index` == old `id` == array position, so `reconcileWipAssignments` keys are unchanged → a WIP written by the OLD build restores correctly on the new build. (No other persisted game-state saves exist.)

### types.ts deletion complete
No dangling `../model/types` import, no `PartyKey`/`ALL_PARTIES`/`PARTY_COLORS`/`scenarioToSpike` reference in src or tests. All 4 BUILD files retarget `types_lib`→`runtime_lib`+`party_lib` with correct new deps; no `types_lib` target survives. (Remaining `*types*` hits are unrelated `spec_types_lib`/`runtime_types_lib`.)

### Adversarial angles (≥2 not on the author's list) + what I could NOT verify
1. **Detail-string keyspace change** — posted inline (evaluate.ts:288). Real user-visible output delta, untested; note it in GAME-112.
2. **scenario-009 cat/dog under the *old* build's `partyIdToKey`** — confirmed 2-party so old first→R/second→D mapping and new direct-keying yield identical results; no divergence.
3. **6-months-drift**: `PARTY_PALETTE`/`PARTY_LABELS` are 5 entries and `partyColor`/`partyLabel` fall back to grey/`String(id)` past index 5 — a future 6th party degrades gracefully, doesn't throw. Fine.
4. **Could NOT verify by observation** (verified only algebraically): rendered lean-map fill, result-card layout, and CSS-var colors — no screenshot/visual diff was run. And N≥3-party correctness is explicitly deferred to GAME-112; I only confirmed no such scenario ships today, so it's out of scope here.

No scope creep: no metric-normalization, turnout/eligibility, or `Party.color` changes snuck in — pure structural unification as intended.

game/web/src/simulation/evaluate.ts
cgruber: Confirmed and tracked — thanks for catching this. It's a lateral change (raw `R`/`D` internal key → raw scenario party-id), not a regression, and pass/fail logic is untouched. Deliberately leaving it as-is in this structural-unification PR to avoid scope creep. Recorded in GAME-112's Current State with a pointer to GAME-113 (framing copy): when the metrics go multiparty, these `detail` strings should interpolate the scenario's authored party *name*/abbreviation rather than the raw id. No change in this PR.

game/web/src/model/runtime.ts
cgruber: Thanks for the assertion-by-assertion audit — this is exactly the behavior-preservation bar this PR needed to clear. Acknowledging the LGTM and the two items you flagged as unverifiable: (1) rendered output was checked only algebraically here — the PuOr lean fill / result-card layout / CSS-var colors get a visual eyeball on the next deploy per the visual-PR process; (2) N≥3-party correctness is explicitly GAME-112, and this PR only claims equivalence for the 12 shipped 2-party scenarios. Merging on this.

PR #320, opened 2026-07-01
feat(pipeline): N-party demographics stage (GAME-116 — resolves)

cgruber:
## Summary

**Resolves GAME-116.** Extends the map-generation pipeline's demographics stage
(`pipeline/demographics-stage.ts` `addDemographics`) from hardcoded two-party to **N parties**, so
multiparty scenarios — the GAME-112 3-party `tutorial-005` demo — can be generated the
"generator = initial state" way instead of hand-authored. Design of record:
`thoughts/shared/plans/2026-06-30-game-116-n-party-demographics.md`.

**Model (the AC's byte-identity constraint forced it).** The current stage is structurally "primary
gets `base + jitter` (clamped); the other party divides the leftover." A renormalizing model would
perturb the 2-party numbers, so the N-party generalization keeps that shape:

1. `primaryShare = clamp(zone.party_base[parties[0]] + jitter, 0, 1)` — the SAME single seeded jitter
   draw, in the SAME order (before the turnout draw); no new PRNG draws → determinism preserved.
2. `remainder = 1 − primaryShare`, split among the non-primary parties by their `party_base` values
   as **weights** — proportional when any are specified, **equal** when none are.

**Byte-identical for the existing 12 two-party scenarios — by construction.** Existing specs author
only the primary base, so the single other party has weight-sum 0 → gets `remainder / 1 =
(1 − primary) × 1.0`. IEEE-754 multiply-by-exactly-1.0 is exact and `1/1 === 1.0`, so this is
**bit-identical** to the old `secondary = 1 − primary`; same key order → identical JSON. Locked by a
new regression test asserting `vote_shares[secondary] === 1 − vote_shares[primary]` exactly.

**Authoring convention** (documented in code): non-primary `party_base` values are weights over the
remainder, not absolute shares — author all N to sum to ~1.0 (with the primary) for the numbers to be
realized literally (e.g. `{ken: 0.55, ryu: 0.37, ind: 0.08}`); the primary alone carries the jitter.

- No `spec-types.ts` / assembler change — `DemographicsSpec.parties`/`ZoneSpec.party_base` are already
  N-capable, and GAME-043 already carries N parties + `Party.color` through assembly.

- The static web game's generation pipeline (`game/web/src/pipeline`) — offline scenario-authoring
  tooling. No runtime/gameplay change; existing scenario JSONs are unchanged and regenerate identically.

## Validation performed

- [x] `bazel test //game/...` green — **47/47** (incl. Biome lint gate). New N-party tests: 3-party
  shares include all parties and sum to 1.0; authored bases realized at jitter 0; unspecified
  non-primary others split the remainder equally; N=3 determinism; and the 2-party
  `secondary === 1 − primary` byte-identity regression guard.
- [x] Byte-identity of existing 2-party regeneration is guaranteed by construction (the N=2 path
  reduces to `remainder × 1.0`, bit-identical to the old arithmetic, PRNG order unchanged) and locked
  by the regression test. (Note: a full-scenario regen-diff is confounded by pre-existing per-scenario
  hand-tweaks/overlays — e.g. the `gen-tutorial-003` ash/birch↔ken/ryu divergence noted in #319 — so
  the faithful guarantee is at the stage's `vote_shares` output level, which the tests pin.)

## Deployment steps

- None — offline generation tooling; ships with the next deploy with no behavior change.

## Tickets addressed

- **Resolves GAME-116.** TICKETS.md updated (Open → Resolved) and the ticket file removed in this branch.

## Rollback

- Revert this commit. Restores the 2-party-only `addDemographics`. No data/schema change; existing
  scenario JSONs are untouched either way.


game/web/src/pipeline/demographics-stage.ts
```diff
@@ -29,10 +45,21 @@ export function addDemographics(partial: PartialScenario, spec: DemographicsSpec
...
 		const primaryShare = Math.min(1, Math.max(0, primaryBase + rawJitter));
-		const secondaryShare = 1 - primaryShare;
+		const remainder = 1 - primaryShare;
+
+		const weights = otherParties.map((op) => zone.party_base[op] ?? 0);
+		const weightSum = weights.reduce((a, b) => a + b, 0);
+
+		const vote_shares = { [primaryParty]: primaryShare } as Record<PartyId, number>;
+		otherParties.forEach((op, i) => {
+			const w = weightSum > 0 ? weights[i]! / weightSum : 1 / otherParties.length;
```
cgruber, line 60: NIT (non-blocking, forward-looking — no shipped scenario is affected): explicit-zero non-primary weights collapse to equal-split, not to zero.

`weightSum > 0 ? weights[i]! / weightSum : 1 / otherParties.length`

If a future 3-party spec authors all non-primary bases as explicit `0` — e.g. `party_base: { ken: 0.55, ryu: 0, ind: 0 }` — then `weightSum === 0` takes the equal-split branch, and ryu/ind each get `remainder/2` rather than the authored `0`. An author writing explicit zeros to mean "these parties get nothing" would be surprised. This can't cause a shipped regression today (the only specs that route through this stage — the 5 `.spec.yaml` files — are all primary-only 2-party; the 3-party tutorial-005 is future GAME-112), so it's a latent authoring footgun, not a bug in this PR. Worth a one-line comment noting "all-zero non-primary weights fall back to equal split" if 3-party specs land.

game/web/src/pipeline/demographics-stage.ts
```diff
@@ -16,10 +16,26 @@ function matchesFilter(filter: ZoneFilter, q: number, r: number): boolean {
...
+ *
+ * `party_base` for a non-primary party is a WEIGHT over the remainder, NOT an
+ * absolute share: author all N bases to sum to ~1.0 (with the primary) for the
+ * numbers to be realized literally (e.g. `{ken: 0.55, ryu: 0.37, ind: 0.08}`).
+ * The primary party alone carries the seeded jitter.
+ *
+ * Byte-identical for the 2-party case: with only the primary base specified, the
+ * single other party has weightSum 0 → gets `remainder / 1 = (1 − primary) × 1.0`,
+ * which is bit-identical to the pre-GAME-116 `secondary = 1 − primary`.
+ */
```
cgruber, line 33: Byte-identity claim VERIFIED (independent adversarial review). Confirming the load-bearing claim in this comment holds, from primary sources + empirically:

1. All specs that route through `addDemographics` author only the PRIMARY base. The only real caller is `pipeline-runner.ts`, driven by the 5 `*.spec.yaml` files (scenario-002, tutorial-001..004). Every `party_base` in those files is `{ ken: N }` (primary-only) — none author `ryu`. So every zone hits `weightSum === 0` → equal-split branch → `1 / otherParties.length` = `1/1` = `1.0` → `remainder * 1.0`. (scenario-003..009 are Kotlin `.kts`-generated and never route through this stage, so the "12 scenarios" set is trivially untouched for the other 7.)

2. IEEE-754: `x * 1.0 === x` for all finite doubles; `1/1 === 1.0` exactly. Empirically confirmed bit-identical across 10,006 samples (real mulberry32 jitter draws over the actual `{ken:0.55}`+jitter±0.04 shape, plus FP edge cases incl. clamped 0/1 and subnormal-adjacent primaries): 0 mismatches by raw Float64 bit pattern.

3. PRNG draw order unchanged — exactly one `nextDouble` jitter draw then one turnout draw per precinct, no added/reordered draws. Key insertion order unchanged (primary then spec-order others = ken, ryu); both keys non-numeric so JSON emits them in insertion order.

The strict-`===` regression test (`demographics-stage_test.ts:183`) exercises the real primary-only 2-party shape and would catch a 1-ULP drift.

COULD NOT VERIFY (named, non-blocking): there is NO golden-file regen guard in CI — `game/scenarios/BUILD.bazel` only wraps the committed JSONs in a filegroup/js_library; nothing regenerates and diffs them against the committed output. So byte-identity is proven analytically + by the standalone bit-identity check above, NOT by an automated regenerate-and-diff. If a masked numeric regression ever were introduced here, CI would not catch it. Consider a future `diff_test` that regenerates each `.spec.yaml` and asserts equality with the committed `.json`.

game/web/src/pipeline/demographics-stage.ts
```diff
@@ -29,10 +45,21 @@ export function addDemographics(partial: PartialScenario, spec: DemographicsSpec
...
-		const secondaryShare = 1 - primaryShare;
+		const remainder = 1 - primaryShare;
+
+		const weights = otherParties.map((op) => zone.party_base[op] ?? 0);
+		const weightSum = weights.reduce((a, b) => a + b, 0);
+
+		const vote_shares = { [primaryParty]: primaryShare } as Record<PartyId, number>;
+		otherParties.forEach((op, i) => {
+			const w = weightSum > 0 ? weights[i]! / weightSum : 1 / otherParties.length;
+			vote_shares[op] = remainder * w;
```
cgruber, line 61: NIT (non-blocking, attack-the-validation): this N=3 sum test uses `Math.abs(sum - 1.0) < 1e-10`, i.e. it asserts sum-to-1 only within FP tolerance. At N=2 the sum is EXACT (primary + remainder*1.0 = primary + (1-primary) = 1 bit-exactly, per the byte-identity path). But for N>=3, `Σ remainder*(wᵢ/weightSum)` = `remainder * (Σwᵢ/weightSum)` where `Σwᵢ/weightSum` is only 1.0 to within rounding — so the emitted shares sum to 1.0 ± a few ULP, not exactly. That's fine for a probability-share model, but if any downstream consumer ever assumes exact-sum-to-1 for N-party scenarios, this test cannot see that failure (it's tolerance-blind by construction). No 3-party scenario ships in this PR, so this is forward-looking only. Model correctness itself is fine: all keys present, each share ∈ [0, remainder] ⊆ [0,1], deterministic (verified — mulberry32, no Math.random/Date.now, no added draws).

game/web/src/pipeline/demographics-stage.ts
cgruber: Acknowledged — real latent footgun, no shipped scenario affected. An all-zero non-primary weight vector (`{ken:.55, ryu:0, ind:0}`) has `weightSum === 0` and so hits the equal-split branch rather than giving each a literal 0. The documented convention (non-primary bases are weights over the remainder; author all N to sum ~1) steers away from it, and a 0 that means "this party doesn't run here" is better expressed by omitting the party from the zone. Leaving the behavior as-is to keep this PR a focused stage change; an explicit "all-zero non-primary ⇒ shares are 0, primary absorbs" guard is a reasonable follow-up if authoring ever needs it. No change here.

game/web/src/pipeline/demographics-stage.ts
cgruber: Thanks — the independent empirical bit-identity check (10k+ real mulberry32 samples over the actual `{ken:0.55}`±jitter shape + FP edge cases, zero Float64 mismatches) is exactly the confirmation the byte-identity claim needed. Noted your point that it's proven analytically + by that standalone check, not by an automated regen guard in CI — I've filed that CI golden-file regen `diff_test` as a separate forward-looking task so a future generator change can't silently drift the shipped JSONs.

game/web/src/pipeline/demographics-stage.ts
cgruber: Correct, and intended: at N≥3 the remainder split is `remainder × (w/weightSum)`, so the sum can only be exact within FP — a strict `===1` there would be wrong. The tolerance is the right assertion for the N≥3 case. The EXACT sum-to-1 property that matters for regression — the 2-party path — is locked separately and strictly by the `vote_shares[secondary] === 1 − vote_shares[primary]` test (no tolerance). So the two tests are deliberately split: exact where it must be, tolerant where FP forces it. No change.

PR #321, opened 2026-07-01
feat(campaigns): debug-only campaign gating (GAME-115 — resolves)

cgruber:
## Summary

**Resolves GAME-115.** Adds the debug-only-campaign **gating harness** so test/demonstration
scenarios can ship in the bundle but stay hidden from normal players — the home for the GAME-112
3-party `tutorial-005` demo. This PR is the gating mechanism only; the demo scenario itself is
GAME-112's deliverable (the debug campaign points at `tutorial-005`, which lands there).

- **`model/campaigns.ts`:** `debugOnly?: boolean` on the `Campaign` interface + a pure
  `visibleCampaigns(isDebug: boolean)` helper that filters `debugOnly` campaigns out unless debug is
  active. Added a `debug` campaign ("Debug (dev)", `debugOnly: true`) pointing at `tutorial-005`.
- **`main.ts`:** the campaign-select render loop iterates `visibleCampaigns(IS_DEBUG)` instead of the
  raw registry (`IS_DEBUG` is the existing sticky-via-sessionStorage debug flag — no propagation work
  needed). Added `tutorial-005` to `CAMPAIGN_ONLY_SCENARIOS` so it never appears in the all-scenarios
  fallback list.
- **Design note:** the gating logic was extracted to a pure helper (`visibleCampaigns`) rather than
  inlined in the untested `showCampaignSelect` DOM code, so it gets fast deterministic unit coverage
  (aligns with the quality-review's "extract pure logic from the main.ts god-file" theme).

The debug campaign references `tutorial-005` before it exists; the card renders and is gated
correctly, but clicking it before GAME-112 lands the scenario would fail to load — acceptable for an
explicitly debug-gated, in-progress feature (only developers with `&debug` can reach it).

- The static web game's campaign-select UI (`game/web` — model/campaigns + main). Gating-only,
  additive; shipped players see no change (two campaigns, exactly as before).

## Validation performed

- [x] `bazel test //game/...` green — **47/47** (incl. Biome lint gate). New coverage:
  - Unit (`campaigns_test.ts`): `debug` campaign flagged `debugOnly`; tutorial/educational are not;
    `visibleCampaigns(false)` hides debug (2 campaigns), `visibleCampaigns(true)` includes it (3);
    non-debug campaigns always visible; registry length now 3.
  - e2e (`campaign-select.spec.ts`): `?view=campaigns` shows 2 cards and NOT "Debug (dev)";
    `?view=campaigns&debug` shows 3 cards including "Debug (dev)".

## Deployment steps

- None — ships with the next deploy; players see no change (the debug campaign is gated off by default).

## Tickets addressed

- **Resolves GAME-115.** TICKETS.md updated (Open → Resolved) and the ticket file removed in this branch.

## Rollback

- Revert this commit. Removes `debugOnly`/`visibleCampaigns`, the debug campaign, and the
  `CAMPAIGN_ONLY_SCENARIOS` entry. No data/schema change; no effect on shipped campaigns.


game/web/src/main.ts
```diff
@@ -398,7 +402,8 @@ const IS_DEBUG =
 		const cardsEl = document.getElementById("campaign-cards");
 		if (cardsEl) {
 			cardsEl.innerHTML = "";
-			for (const campaign of CAMPAIGN_REGISTRY) {
+			// GAME-115: debug-only campaigns appear only when debug mode is active.
+			for (const campaign of visibleCampaigns(IS_DEBUG)) {
```
cgruber, line 406: Non-blocking (low severity, latent-consistency note — not a leak today).

This render loop was correctly switched to the new `visibleCampaigns(IS_DEBUG)` gate. But `buildContinueUrl` (main.ts:389-396, just above) is a second campaign-enumeration site that still iterates the **raw** `CAMPAIGN_REGISTRY`, so the two campaign-listing paths are now asymmetric: this one is gated, that one isn't.

Why it's *not* a leak in this PR: `buildContinueUrl` only ever emits a `?campaign=debug` URL when `loadLastPlayedScenario()` returns `tutorial-005`, and a normal (non-debug) player has no discoverable way to put `tutorial-005` into `lastPlayed` — the scenario is absent from `SCENARIO_MANIFEST` (the all-scenarios fallback, main.ts:274) and only reachable by manually typing `?campaign=debug`, the documented dev-only escape hatch. So the Continue button can only point at the debug campaign for someone who already used debug mode. Verified — no action needed for GAME-115's gating claim.

Worth a maintainer's awareness for the "six months from now" case: if a future debug scenario ever becomes reachable/completable by a normal player through some other path, this raw iteration would hand them a live `?campaign=debug` deep-link via the Continue button, bypassing the gate. A one-line guard in `buildContinueUrl` (skip `debugOnly` campaigns when `!IS_DEBUG`, mirroring this loop) would make gating uniform across both enumeration sites. Optional — flagging, not requesting.

cgruber: Confirmed and agreed it's not a leak today — thanks for tracing it. The reason it's structurally safe: `buildContinueUrl` only ever runs on a scenarioId already in `lastPlayed`, and a normal (non-debug) player can never reach or play `tutorial-005`, so it can't land in `lastPlayed` for them. For a debug player who did play it, resolving it back to `campaign=debug` is the correct continue URL. So the raw-registry iteration here is behaviorally fine.

Leaving it as-is to keep this PR a focused gating change rather than expanding a non-bug. The symmetry hardening (have `buildContinueUrl` also respect gating, or key off `visibleCampaigns`) is a reasonable future guard once a differently-reachable debug scenario exists — I'll fold that consideration into the GAME-112 demo work, where the debug scenario actually becomes playable. No change in this PR.

PR #322, opened 2026-07-02
feat(metrics): two-party-normalize efficiency-gap + mean-median (GAME-112 PR 1 of 2)

cgruber:
## Summary

**Advances GAME-112 (PR 1 of 2 — the metrics half; ticket stays open for the demo).**
Two-party-normalizes the efficiency-gap and mean-median fairness metrics so a nonzero third party
can't corrupt them — the correctness prerequisite for shipping multiparty scenarios. Sourced from the
2026-06-30 metrics audit. PR 2 is the collaborative 3-party `tutorial-005` demo (generated via
GAME-116, housed in GAME-115's debug campaign).

- **`efficiency_gap` (`evaluate.ts`):** the win line is now half the **two-party** vote
  (`(party1+party2)/2`) rather than half of *all* votes, and the denominator is the sum of two-party
  votes rather than all votes. Third-party ballots are excluded from the two-party contest; a
  third-party seat win still wastes both majors' votes.
- **`mean_median`:** each district's share is now `c.party / (party1 + party2)` — the party's
  **two-party** share — rather than its raw all-vote share.
- **Behavior-preserving at third-party-share 0:** both reduce exactly to the prior formulas when
  `party1 + party2 == total` (all shipped scenarios are two-party), so the existing threshold tests
  are unchanged and remain the regression guard.

- The static web game's simulation layer (`game/web/src/simulation/evaluate.ts`). Metric-math change;
  no rendering/store/schema change. Two-party scenarios are unaffected.

## Validation performed

- [x] `bazel test //game/...` green — **47/47** (incl. Biome lint + winnability e2e). The existing
  two-party `efficiency_gap`/`mean_median` threshold tests still pass unchanged (behavior-preserving
  guard); the winnability e2e confirms no shipped scenario's metric crossed a threshold.
- [x] New 3-party `evaluate_test` fixtures actively prove the normalization:
  - efficiency_gap: the two-party gap is **0.083** (`≤0.09` passes, `≤0.08` fails) — the OLD all-vote
    formula would give **0.10**, which `≤0.09` would fail, so the passing test can only hold under
    two-party normalization.
  - mean_median (ken): the two-party diff is **0.083** (`≤0.075` fails, `≤0.09` passes) — the OLD raw
    formula would give **0.067**, which `≤0.075` would pass, so the failing test can only hold under
    two-party normalization.

## Deployment steps

- None — ships with the next deploy; no behavior change for the (two-party) shipped scenarios.

## Tickets addressed

- **Advances GAME-112 (PR 1 of 2).** Ticket stays open for the 3-party `tutorial-005` demo + a
  3-party adapter test (PR 2). Ticket Current State updated with PR-1 status.

## Rollback

- Revert this commit. Restores the all-vote efficiency-gap denominator/win-line and the raw-share
  mean-median. No data/schema change; two-party scenarios identical either way.

### DIFF
--- a/game/web/src/simulation/evaluate.ts
+++ b/game/web/src/simulation/evaluate.ts
@@ -254,34 +254,44 @@ export function evaluateCriteria(
 			}
 
 			case "efficiency_gap": {
-				// Wasted-vote efficiency gap between the two major parties (the
-				// scenario's first two parties, party1/party2). GAME-043 keeps this
-				// two-party to preserve behavior; the proper two-party normalization
-				// for N>2 is GAME-112.
+				// Two-party-normalized wasted-vote efficiency gap between the two major
+				// parties (the scenario's first two parties, party1/party2). GAME-112:
+				// the win line, the wasted-vote counts, AND the denominator are all
+				// restricted to the two-party vote (party1 + party2), so a nonzero third
+				// party cannot corrupt the gap. When third-party share is 0 this reduces
+				// to the plain two-party EG up to floating-point (party1+party2 == 1), so
+				// the shipped two-party scenarios are preserved to display/threshold
+				// precision; a tight-boundary 2-party regression test (evaluate_test)
+				// guards against a gross formula change.
 				// Positive = party1 has more wasted votes; negative = party2 does.
 				// Criterion uses abs(gap) so direction doesn't matter.
 				const party1 = parties[0]!;
 				const party2 = parties[1]!;
 				let p1Wasted = 0;
 				let p2Wasted = 0;
-				let allVotes = 0;
+				let twoPartyVotes = 0;
 				for (const dr of simResult.districtResults) {
 					const p1Votes = (dr.voteTotals[party1] ?? 0) * dr.totalVotes;
 					const p2Votes = (dr.voteTotals[party2] ?? 0) * dr.totalVotes;
-					allVotes += dr.totalVotes;
+					// Half the TWO-PARTY vote is the win line, not half of all votes —
+					// third-party ballots are excluded from the two-party contest.
+					const twoPartyTotal = p1Votes + p2Votes;
+					const winLine = twoPartyTotal * 0.5;
+					twoPartyVotes += twoPartyTotal;
 					if (dr.winner === party1) {
-						p1Wasted += Math.max(0, p1Votes - dr.totalVotes * 0.5);
+						p1Wasted += Math.max(0, p1Votes - winLine);
 						p2Wasted += p2Votes;
 					} else if (dr.winner === party2) {
-						p2Wasted += Math.max(0, p2Votes - dr.totalVotes * 0.5);
+						p2Wasted += Math.max(0, p2Votes - winLine);
 						p1Wasted += p1Votes;
 					} else {
-						// Third-party winner: both major parties wasted all votes
+						// Third-party winner: both major parties lost the seat → all their
+						// two-party votes are wasted.
 						p1Wasted += p1Votes;
 						p2Wasted += p2Votes;
 					}
 				}
-				const rawGap = allVotes > 0 ? (p1Wasted - p2Wasted) / allVotes : 0;
+				const rawGap = twoPartyVotes > 0 ? (p1Wasted - p2Wasted) / twoPartyVotes : 0;
 				const absGap = Math.abs(rawGap);
 				passed = applyOp(absGap, c.operator, c.threshold);
 				const direction = rawGap >= 0 ? `${party1}-disadvantaged` : `${party2}-disadvantaged`;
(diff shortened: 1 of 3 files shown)


game/web/src/simulation/evaluate.ts
```diff
@@ -290,13 +298,21 @@ export function evaluateCriteria(
...
 				// Large negative value = party wins more seats than votes warrant (cracked opponents).
 				// Criterion applies applyOp to the raw (signed) difference.
-				const shares = simResult.districtResults.map(
-					(dr) => (dr.voteTotals as unknown as Record<string, number>)[c.party] ?? 0,
-				);
+				const party1 = parties[0]!;
+				const party2 = parties[1]!;
+				const shares = simResult.districtResults.map((dr) => {
+					const pShare = (dr.voteTotals as unknown as Record<string, number>)[c.party] ?? 0;
+					const twoParty = (dr.voteTotals[party1] ?? 0) + (dr.voteTotals[party2] ?? 0);
```
cgruber, line 321: **Latent regression: `mean_median` on a non-major target party now divides by a denominator that excludes it (can exceed 1).**

New per-district share is `c.party / (party1 + party2)` (line 313 sums only `parties[0]` + `parties[1]`). When `c.party` is one of the two majors this is correct — it's the target's two-party share. But if a scenario ever authors `mean_median` with `party` set to a **third** party, `c.party` is not in the denominator, so you compute e.g. `ind / (ken + ryu)`, which is unbounded above 1 and is not a vote share of anything meaningful. The OLD code returned `voteTotals[c.party]` — the raw all-party share — which is always a valid 0..1 quantity for any target. So for a third-party target this change is a *regression* in correctness, not an improvement.

**Not live today:** I checked all four shipped scenarios that use these metrics (003/004/007/008). Only scenario-004 authors `mean_median`, and it targets `ken` = `parties[0]` (a major), so nothing is broken now. This is latent coupling that will bite the moment PR 2 (or GAME-116's N-party stage) authors a `mean_median` on a minor party.

Recommendation: either (a) normalize by the two-party pair that *includes* `c.party` when it's a major and fall back to raw share otherwise, or (b) add an explicit guard/assert that `c.party ∈ {party1, party2}` for `mean_median`, and document the restriction. A one-line note here would also help the PR-2 author avoid the trap.

Could-not-verify: I could not confirm PR 2's planned 3-party scenario won't target a minor party with `mean_median` — flagging so it's on the record before that scenario lands.

game/web/src/simulation/evaluate.ts
```diff
@@ -254,34 +254,42 @@ export function evaluateCriteria(
...
-				// scenario's first two parties, party1/party2). GAME-043 keeps this
-				// two-party to preserve behavior; the proper two-party normalization
-				// for N>2 is GAME-112.
+				// Two-party-normalized wasted-vote efficiency gap between the two major
+				// parties (the scenario's first two parties, party1/party2). GAME-112:
+				// the win line, the wasted-vote counts, AND the denominator are all
+				// restricted to the two-party vote (party1 + party2), so a nonzero third
+				// party cannot corrupt the gap. When third-party share is 0 this reduces
+				// exactly to the plain two-party EG (behavior-preserving for the shipped
+				// two-party scenarios — the existing threshold tests are the guard).
```
cgruber: **The existing 2-party threshold tests do NOT actually guard behavior-preservation — they can only catch gross regressions.**

The comments (evaluate.ts:263, :305) and the ticket both lean on "the existing threshold tests are the guard" for the behavior-preserving claim. Attacking that validation: those tests are threshold pass/fail assertions sitting far from their boundaries, so they structurally cannot observe a sub-threshold shift:
- 2-party EG fixture value ≈ 0.033, asserted `lte 0.05` (pass) and `lte 0.02` (fail) — nearest boundary 0.017 away.
- 2-party mean_median: balanced diff = 0 vs 0.05, gerrymandered diff ≈ 0.167 vs 0.05 — nearest boundary 0.05 away.

No test anywhere makes an exact old-vs-new value assertion (`assertEqual` on the metric). So "behavior-preserving" is asserted, not pinned. In practice it *is* preserved to ~1e-16 — I verified that at evaluate time (`election.ts` re-derives `voteTotals` by dividing each party's population-weighted sum by `totalPop`, with NO rounding, and every shipped 2-party scenario's group `vote_shares` sum to exactly 1 over {ken,ryu}), so `party1 + party2 = 1` up to pure FP and the FP delta is below `toFixed(1)` and every threshold margin. The claim is true; the tests just aren't what makes it true.

Not blocking (the property holds), but the "the existing threshold tests are the guard" framing overstates what those tests do. If you want a real regression guard for behavior-preservation, an exact-value assertion on the 2-party EG/mean_median fixtures (e.g. `assertEqual`-with-epsilon comparing the metric to a hard-coded expected) is the only thing that would catch a future structural break in this reduction.

game/web/src/simulation/evaluate.ts
```diff
@@ -254,34 +254,42 @@ export function evaluateCriteria(
...
+					twoPartyVotes += twoPartyTotal;
 					if (dr.winner === party1) {
-						p1Wasted += Math.max(0, p1Votes - dr.totalVotes * 0.5);
+						p1Wasted += Math.max(0, p1Votes - winLine);
 						p2Wasted += p2Votes;
 					} else if (dr.winner === party2) {
-						p2Wasted += Math.max(0, p2Votes - dr.totalVotes * 0.5);
+						p2Wasted += Math.max(0, p2Votes - winLine);
 						p1Wasted += p1Votes;
 					} else {
```
cgruber, line 287: **The third-party-WIN branch (lines 285-290) ships an unguarded, untested convention.**

When a seat is won by neither major (a third party takes the plurality), the code wastes BOTH majors' full two-party votes. That is a *modeling choice* — the standard wasted-vote EG is defined for two-party (single-winner between the two) contests and is genuinely undefined when the winner is outside the pair — not a canonical formula, so it's worth being explicit that it's a choice.

Coverage gap: the 3-party fixture keeps `ind` at 0.2 in every district, so `ind` is always third and this branch NEVER executes in tests. The branch is dead under the current suite. Per the ticket this is deferred to PR 2 (a 3-party adapter test where "a district a third party wins"), which is the right place — flagging only so it's on record that this PR ships the convention without any test exercising it, and that whether "both majors waste all their votes" is the desired pedagogy for a third-party-won seat is a design decision PR 2 should confirm (it inflates the gap toward whichever major had more votes in that seat).

Not blocking given the explicit PR-2 deferral.

game/web/src/simulation/evaluate.ts
cgruber: Fixed. Added a `targetIsMajor = c.party === party1 || c.party === party2` guard: a non-major target now falls back to the raw share (two-party normalization is only meaningful for a major party; dividing by `party1+party2` when it excludes `c.party` blows past 1). Added a regression test (`mean_median` minor target `ind`, raw shares [0.6,0.2,0.2] → guarded diff 0.133 ≤ 0.20 passes; the unguarded two-party value would be 0.417 and fail) so the fallback is pinned. Thanks — this was a real latent footgun I introduced.

game/web/src/simulation/evaluate.ts
cgruber: Fixed. You're right — the pre-existing threshold tests sit far from their boundaries and there was no exact-value assertion, so they only caught gross regressions. I (1) softened the two overstated comments to "preserved to display/threshold precision" (not "the threshold tests are the guard"), and (2) added tight-boundary 2-party regression tests that actually pin the values: efficiency-gap on EG_PRECINCTS to (0.033, 0.034] (`lte 0.034` passes, `lte 0.033` fails) and mean-median on MM_GERRYMANDER to (0.16, 0.17] (`lte 0.17` passes, `lte 0.16` fails). A formula change that reverted the two-party normalization would shift those values out of the pinned band and fail these.

game/web/src/simulation/evaluate.ts
cgruber: Acknowledged — no change in this PR. The third-party-WIN branch ("both majors waste all their two-party votes") is an intentional convention but, as you note, it has no test coverage because no shipped scenario has a district a third party wins. This is correctly deferred to GAME-112 PR 2, where the 3-party `tutorial-005` demo (a real scenario a third party can contest) will exercise this path with coverage. It's recorded in the ticket's PR-2 remaining-work list. Flagging-for-the-record accepted.

PR #324, opened 2026-07-03
docs(plan): candidate/independent model + file GAME-117..121

cgruber:
## Summary

Design of record for the next arc — the **"candidates run in districts" model** — plus the five
follow-on tickets it decomposes into. Approved with the user 2026-07-02. Docs/planning only, no code.

The idea: a district seat is won by a **named candidate**. A **party** fields a candidate in every
district (can win many seats); an **independent** is a candidate in **one** district — their home — and
wins ≤ 1 seat. Both the "outcomes read as people" ask and the independent mechanic fall out of one
model. Scenarios without an independent are unchanged; candidate names are optional (fall back to the
party name), so nothing existing shifts.

- **`plans/2026-07-02-district-candidates-and-independents.md`** (+ `.compressed.md`) — the model, the
  lean-vs-ballot distinction, per-ticket scope, and sequence.
- **Tickets filed (TICKETS.md Open):** GAME-117 (named per-district candidates — keystone), GAME-118
  (home-base independent, opt-in), GAME-119 (feature-anchored zone filters), GAME-120 (rebuild
  tutorial-005 as the multi-party tutorial), GAME-121 (independent tutorial). Sequence: 117 → 118 → 119
  (features) → 120, 121 (tutorials).

- None (documentation + tickets under `thoughts/`). No build/runtime impact.

## Validation performed

- [x] Design doc has a `.compressed.md` sibling (two-form convention).
- [x] TICKETS.md updated in the same change (5 Open rows); IDs 117–121 are new (116 was the prior high).
- [ ] N/A — no code, no tests, no deploy.

## Deployment steps

- None — docs only.

## Tickets addressed

- Files GAME-117..121 (planning). None resolved here.

## Rollback

- Revert this commit — removes the design doc + the five ticket files + their TICKETS.md rows.


thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md
```diff
@@ -0,0 +1,99 @@
...
+- **Out of scope:** the independent restriction (that's 118); this is names only.
+
+## GAME-118 — home-base independent candidates (opt-in)
+
+- **Model:** an independent is a candidate with `independent: true` and a **home** (a precinct id, or
+  a coordinate resolved to a precinct). At election time:
+  - In the district containing the home → the independent is on the ballot; district winner = plurality
+    over parties **+ the independent**.
+  - In every other district → the independent is **excluded from the seat contest**; the winner is the
+    plurality over the **parties only** (his supporters there fall back to a party for the seat — his
```
cgruber, line 53: **Modeling gap (core semantics, not a schema detail): what is the STATUS of an independent's votes outside his home district?** This is the load-bearing question the design of record must pin down, and right now it's glossed.

The AC says non-home winner = "plurality over the parties only" — that rule itself is fine. The problem is the parenthetical justification: *"his supporters fall back to a party for the seat."* That phrasing implies a **vote transfer** (second-preference reallocation) that the current data model **cannot express**. Since GAME-116, each precinct carries a fixed `share` per party (dhalsim is one of those parties). There is no second-preference field and no rule for *which* party an excluded independent's share flows to, or *how much*. So "fall back to a party" is not something the model does — the independent's share simply **isn't counted** in the seat tally.

Recommend the design state this precisely, because it settles two things at once:
1. **Winner:** the independent's share is *excluded from the seat tally* (not transferred); whoever leads **among the parties** wins. Say "excluded," not "falls back."
2. **Consequence to own in the UX bullet:** a party can then win a district where the independent held the *raw plurality* (IND 0.45 / P1 0.30 / P2 0.25 → P1 wins). That is the "independent with a plurality in a non-home district" case. Pedagogically rich, but it reads as "the candidate with the most votes lost" — a real legibility risk. The non-home "party-only race" note must make this legible, or players will read it as a bug.

The same question also decides the metrics denominator (is the excluded share in or out of Σ votes?) — see separate note. One sentence in the design ("outside home, the independent's votes are dropped from the seat tally and from the two-party metric denominator") pins the winner, the UX, and the metrics together. Deferring *this* to the ticket (the way schema SHAPE is fairly deferred) isn't acceptable for a design-of-record — it's the core semantics of the mechanic.

thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md
```diff
@@ -0,0 +1,99 @@
...
+
+## GAME-121 — independent tutorial
+
+- A new scenario built on the home-base independent (118) + candidates (117) — e.g. Dhalsim done right:
+  one home district, a base you can pack or crack, a home pin. Teaches the independent mechanic
+  specifically, distinct from the multi-party tutorial.
+
+## Sequence & rationale
+
+117 → 118 → 119 (features), then 120, 121 (the tutorials). 117 is the keystone: it makes outcomes read
```
cgruber, line 106: **Dependency review — the DAG is clean (no cycles), with one nuance worth stating.** I traced the stated dependencies:

- 117 → blocks {118, 120, 121}
- 119 → feeds {120, 121}
- 118 → feeds 121

That is a valid DAG (topological order 117 → 118/119 → 120 → 121); no circular dependency. 116 was the prior high ticket and 117–121 are new — no ID collision, including the Resolved section.

Nuance for sequencing flexibility: the dependencies are not all equally hard.
- **HARD:** 118 → 117 (the independent *is* a candidate-in-one-district, so the candidate model must exist first) and 121 → 118 (the tutorial needs the engine mechanic).
- **SOFT:** 120 → 117 and 121 → 117. Candidates are **display-only** (per GAME-117's own "no change to who WINS" AC) — a tutorial can ship functionally with party-name fallback and get candidate names later. So if 117 slips, 120 can still proceed on party names.

Optional: note the hard-vs-soft split in §SEQUENCE so a future scheduler knows 117 isn't a hard blocker for the *tutorials*, only for 118.

thoughts/shared/tickets/GAME-118-home-base-independent.md
```diff
@@ -0,0 +1,50 @@
...
+
+- GAME-112 shipped third-party support where a "party" contests every district and can win multiple
+  seats. An independent authored as a party (the tutorial-005 draft's Dhalsim) wrongly behaves the same
+  — he could win any district he leads, and is "on the ballot" everywhere.
+- `election.ts` computes each district's winner as the plurality over all `parties` (GAME-043 runtime).
+- The two-party metric normalization (GAME-112 PR1) already restricts EG/mean-median to the two majors.
+
+## Goals / Acceptance Criteria
+
+- [ ] A party may be flagged `independent: true` with a **home** (precinct id or coordinate → precinct).
```
cgruber, line 26: **Blind spot: multiplicity is never specified.** The plan and this ticket model "an independent" (singular) with "a home." Two under-specified cases an implementer will hit:

1. **N independents.** May a scenario author more than one `independent: true` party, each with its own home? The election rule generalizes cleanly (each is on the ballot only in its own home district, excluded elsewhere), but nothing states it — so the implementer has to guess whether to loop or assume one. If v1 is intentionally single-independent, say so as an explicit scope boundary; if N is allowed, the schema (list of independents) and the "excluded elsewhere" rule need to name it.

2. **Two independents home-in the SAME district.** If N is allowed, the home district's ballot is now `parties + independent_A + independent_B` — fine, but worth one line so it isn't discovered as a surprise.

Pick one and write it down; either answer is cheap, but silence forces a guess in the exact ticket (GAME-118) the design is supposed to fully specify.

thoughts/shared/tickets/GAME-121-independent-tutorial.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Current State
+
+- No independent-specific scenario exists; GAME-118 delivers the engine model + the home pin.
+- The multiparty rendering (all-party card, plurality lean, breakdown) is already shipped.
+
+## Goals / Acceptance Criteria
+
+- [ ] A scenario with one home-base independent (GAME-118): a home precinct + a concentrated regional
+      base, two major parties, feature-anchored leans (GAME-119).
```
cgruber, line 29: **Scenario identity is missing for GAME-121.** GAME-120 explicitly *repurposes* `tutorial-005` into the multi-party tutorial ("Rebuild the tutorial-005 draft"). That means the independent tutorial here needs a **different scenario slot** — but no ID is given (tutorial-006? something else?). An implementer picking up GAME-121 cold cannot tell where this scenario lives, and both 120 and 121 currently trace back to the same tutorial-005 draft as their starting point.

Add the target scenario id (and confirm it doesn't collide with an existing/planned tutorial). One line in the AC: "New scenario `tutorial-00X` (not tutorial-005, which GAME-120 claims)." This also clarifies for the reader that 120 and 121 produce two distinct tutorials, not two versions of one.

thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md
```diff
@@ -0,0 +1,99 @@
...
+  name), or per-party `candidates: { <districtId>: <name> }`. Optional; loader validates strings.
+  (Decide the exact shape in the ticket; keep it additive + optional.)
+- **Render:** `panels.ts` result card + `mapRenderer.ts` info/result surfaces show the winning
+  candidate name + party (color/badge unchanged). One person can't hold two seats — per-district
+  names make multi-seat party wins read correctly ("Sakura took D1, Chun-Li took D2 for Ryu").
+- **Out of scope:** the independent restriction (that's 118); this is names only.
+
+## GAME-118 — home-base independent candidates (opt-in)
+
+- **Model:** an independent is a candidate with `independent: true` and a **home** (a precinct id, or
```
cgruber, line 48: **Edge case the design doesn't cover: the home precinct is UNASSIGNED mid-draw.** The whole independent mechanic keys off "the district *containing* the home" — but during a draw, precincts begin unassigned and the player assigns them incrementally. Until the home precinct is placed in some district, there is no "home district," so the rule "on the ballot in the district containing the home; excluded everywhere else" has no defined answer for the independent.

Questions an implementer must resolve and the design should state:
- While the home precinct is unassigned, is the independent on **no** ballot (wins 0 seats that draw), or does the result view simply show an incomplete map?
- The tutorials (120/121) gate on `district_count` = "every precinct is assigned," so a *legal* map always has the home assigned — but the live result preview repaints on *every* edit, including illegal intermediate states. So this state is reachable and visible, and the "party-only race" note must not misfire (e.g. claim a party-only race in a district that doesn't yet contain the home).

Recommend one line: "if the home precinct is unassigned, the independent is on no ballot (0 seats) until it is placed." Cheap to state, awkward to discover.

thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md
cgruber: Good catch — this was underspecified. Decided + documented (design doc GAME-118 "clarifications" + the ticket AC): outside the home district the independent's share is **disregarded for the seat** — the winner is the plurality among the parties as-is, with **no redistribution** to a major (we don't model second preferences). His local support is simply unrepresented on that ballot — the honest outcome, and consistent with the two-party metric normalization (GAME-112), which already excludes non-majors. Precinct vote shares are untouched; only the seat contest excludes him.

thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md
cgruber: Thanks for tracing the DAG — agreed it's acyclic (117 → 118; 117+119 → 120; 117+118+119 → 121). No change needed; the sequence 117 → 118 → 119 → 120, 121 respects it.

thoughts/shared/tickets/GAME-118-home-base-independent.md
cgruber: Fixed — the singular framing was a blind spot. Design doc + GAME-118 AC now state the model supports **zero or more** independents, each carrying its own `home` and on the ballot only in the district containing that home (two sharing a home just both appear on that ballot). No global "the independent" assumption.

thoughts/shared/tickets/GAME-121-independent-tutorial.md
cgruber: Fixed — GAME-121 now specifies a NEW scenario `tutorial-006` (GAME-120 keeps/repurposes `tutorial-005` for the multi-party tutorial; the independent tutorial is its own scenario), plus its `CAMPAIGN_ONLY_SCENARIOS` entry + the debug campaign's `scenarioIds`. Good catch — the two tutorials must not collide on the same id.

thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md
cgruber: Fixed — added to the design doc + GAME-118 AC: if the home precinct is unassigned mid-draw, the independent is on **no** ballot (exactly as an unassigned precinct contributes to no seat) and becomes a contender the moment his home is placed into a district. It falls out of the "on the ballot in the district containing the home" rule — no special-case error path needed.

PR #325, opened 2026-07-03
feat(model): named per-district candidates (GAME-117 — resolves)

cgruber:
## Summary

**Resolves GAME-117 — the keystone of the candidate/independent arc.** A district seat now reads as a
**named candidate** (a person), not an abstract party — which also makes multi-seat party wins read
correctly (one person can't hold two seats) and is the foundation for the home-base independent model
(GAME-118: an independent is just a candidate that exists in one district). Design of record:
`thoughts/shared/plans/2026-07-02-district-candidates-and-independents.md`.

- **Schema (optional, additive):** `Party.candidates?: string[]` (`model/scenario.ts` +
  `pipeline/spec-types.ts` + assembler passthrough) — `candidates[districtId − 1]` is that party's
  candidate in district N (ids are 1-based). Loader `parseParty` validates a string array (rejects a
  non-array, or a non-string element).
- **Resolve + thread:** a pure `candidateForDistrict(candidates, districtId)` helper (`model/party.ts`)
  does the 1-based lookup; `main.ts` builds a `partyCandidates` map (mirroring the `partyNames` /
  `partyColors` idiom) threaded to `renderResults`; the winner badge shows the candidate name when
  authored, **falling back to the party name** otherwise (names escaped — GAME-103).
- **Behavior-preserving:** no scenario authors candidates yet → every existing scenario (and all
  2-party scenarios) render exactly as before. Candidates are display-only — who WINS is unchanged.

- The static web game (`game/web`) — model/pipeline/loader + render/panels + main. Additive optional
  field; existing scenarios unchanged.

## Validation performed

- [x] `bazel test //game/...` green — **47/47** (incl. Biome lint + existing e2e = the no-candidates
  regression guard). New tests: `loader_test` (candidates parse when present; absent → undefined;
  non-array rejected; non-string element rejected) + `candidateForDistrict` (1-based indexing; beyond
  roster / no roster / empty roster → undefined).

## Deployment steps

- None — additive optional field; ships with the next deploy with no behavior change.

## Tickets addressed

- **Resolves GAME-117.** TICKETS.md updated (Open → Resolved) and the ticket file removed.

## Rollback

- Revert this commit. Removes `Party.candidates` + the render lookup; results fall back to party names
  (today's behavior). No data/schema migration.


game/web/src/render/panels.ts
```diff
@@ -30,13 +31,20 @@ export function renderResults(
...
+		return name !== undefined ? escapeHtml(name) : labelOf(p);
+	};
 
 	const { districtResults } = state.simulationResult;
 	const html = districtResults
 		.map((r) => {
 			const color = districtColor(r.districtId);
 			const winnerColor = colorOf(r.winner);
-			const winnerLabel = labelOf(r.winner);
+			const winnerLabel = winnerOfDistrict(r.winner, r.districtId);
```
cgruber, line 47: **Nit / test-coverage gap (non-blocking): the candidate-display path this line adds is pinned by no test.**

The pure helper `candidateForDistrict` is unit-tested (values, 1-based, fallback — good), and I verified by inspection that the fallback here is behavior-preserving. But the *wiring* this line introduces — `winnerOfDistrict` resolving a candidate name, `escapeHtml`-ing it, and rendering it into the `.winner-badge` — executes in **zero tests**:

- No `panels_test.ts` exists; the render layer has no unit tests.
- No scenario authors a `candidates` array (I grepped `game/scenarios/` — the only `candidate*` matches in `tutorial-005` are prose in motivation/hint strings, not a `Party.candidates` key), so no e2e ever drives this branch.
- The closest guard, `a11y.spec.ts` tests 1–2, asserts `resultsText.toContain("Ken Party")` — but "Ken Party" also appears in the vote-details line (`${p1Label} …%`), so that assertion is satisfied whether or not the badge renders correctly. It structurally cannot pin the badge text.

Net: the acceptance criterion "a district win with an authored candidate shows the candidate name … escaping verified" is covered by nothing. Not a defect — the code reads correct — but the PR's "render behavior tested" is only true for the pure helper, not the display. A tiny `renderResults` DOM test (jsdom) asserting the badge shows the candidate when authored, the party name when not, and an escaped `<script>`-style name, would close it and lock in the escaping guarantee for GAME-118.

cgruber, line 47: **Scope note (non-blocking): only the result card was updated; the map/precinct surface was not — and the ticket's AC names both.**

GAME-117's acceptance criterion reads: "The result card (`panels.ts`) **and precinct/result surfaces** show the winning candidate name". This PR updates `panels.ts` only. The one place `mapRenderer.ts` renders a winner *name* as text is the precinct-info hover panel (`mapRenderer.ts:1270–1274`), which still shows the party name via `winnerOf(d.voteShare, …)` + `partyNames[…] ?? partyLabel(…)`.

I believe leaving that as-is is *correct*: that panel shows a single precinct's vote **lean** (which party leads the raw vote share here), not a district **seat winner** — and a candidate is elected per district, not per precinct, so a candidate name has no meaning on one precinct's lean. `hexFill` (`:1437`) likewise uses `winnerOf` only to pick a fill color, no text.

So this is a scoping decision, not a bug — but it is undocumented, and the PR claims to *resolve* GAME-117 whose AC literally names "precinct/result surfaces". Please either (a) add one line to the PR/ticket stating the map surface intentionally stays party-lean (candidates are a district-seat concept), or (b) confirm a follow-up covers any seat-winner surface on the map. Right now a reader can't tell whether the second surface was considered and scoped out, or missed.

game/web/src/pipeline/assembler.ts
```diff
@@ -86,6 +86,7 @@ export function assembleScenario(partial: PartialScenario, spec: AssemblySpec):
 		name: ps.name,
 		abbreviation: ps.abbreviation,
 		...(ps.color !== undefined ? { color: ps.color } : {}),
+		...(ps.candidates !== undefined ? { candidates: ps.candidates } : {}),
```
cgruber, line 89: **Minor (non-blocking): the assembler passes `candidates` through untyped.**

This spreads `ps.candidates` straight onto the assembled `Party` with only a presence check — no shape validation (`Array.isArray` / string elements) at assembly time. The loader's `parseParty` does validate, so a malformed **JSON scenario** is still caught. But a malformed **spec** authored through the pipeline (`assembleScenario`) is not caught here; it surfaces only later when the assembled scenario is re-loaded, with a less local error. It mirrors the existing `color` passthrough on the line above, so this is consistent with the current pattern rather than a regression — just flagging that the pipeline trusts spec authors for this field. Fine to leave if that's the established convention.

game/web/src/render/panels.ts
cgruber: Acknowledged — no change in this PR, and a fair "pinned by nothing" note.

The render layer (`render/panels.ts`, `render/mapRenderer.ts`) has no unit-test harness today — there's no jsdom/DOM setup and no `panels_test.ts` — so pinning the badge in a unit test is a harness addition beyond this additive, behavior-preserving change. What *is* pinned:

- `candidateForDistrict` is unit-tested (1-based indexing, beyond-roster/empty → `undefined`, non-string rejected).
- The wiring is behavior-preserving: no scenario authors `candidates` yet, so every existing scenario renders exactly as before (the party-name fallback path).

The candidate-*display* path (badge → escaped candidate name) gets end-to-end coverage imminently in **GAME-120** (tutorial-005 rebuilt as the multi-party tutorial) and **GAME-121** (the independent tutorial), which author scenarios *with* candidates and assert the rendered result card. The broader "untested UI layer" gap is already tracked in the 2026-06-27 codebase-quality review.

cgruber: No change — by design the map has no seat-winner *name* surface to update. Surfacing + justifying the scoping:

The map's only winner-related surfaces are both intentionally **party**-level:
- Precinct **lean** — `mapRenderer.ts:1270` shows `winnerOf(precinct.voteShare)` as the leading *party* ("Lean: Ken Party (52%)").
- `hexFill` plurality coloring — `:1437`, colour only, no text.

Per the plan's lean model ("**Lean** = voter *preference* … stays map-wide for everyone", line 25) that is deliberately a *party* concept: a precinct *leans toward a party*; it does not *elect a person*. Candidates win district **seats**, and the sole seat-winner display is the `panels.ts` result card ("Candidate (Party) +margin") — which this PR updates. That is exactly the "names only" scope (plan line 44): add names to the existing winner display, which is the result card. There is no per-district seat-winner *name* rendered on the map to convert.

Plan **line 41** ("`mapRenderer.ts` info/result surfaces show the winning candidate") overstates this relative to the lean model. I'll reconcile that wording in the **GAME-118** change, which revisits these same render surfaces (it adds the home-base independent's pin + seat logic) and shares this plan as its design doc. A dedicated district-result overlay on the map would be a *new* surface = a follow-up, not part of GAME-117's "names only".

game/web/src/pipeline/assembler.ts
cgruber: No change — this matches the assembler's established idiom for *every* optional field, not just `candidates`:

- `color` (line 88), `candidates` (line 89), plus `name`, `initial_district_id`, `character`, `party_id`, `flavor_text`, … all use the same `...(x !== undefined ? { x } : {})` spread passthrough, untyped at assembly.

The authoritative validation is at **load**: `parseParty` rejects a non-array `candidates` or a non-string element (covered by `loader_test`). `spec-types.ts` also types `PartySpec.candidates?: string[]`, so TS-authored specs are checked at compile time. Adding a generation-time string-array check *here* would duplicate the loader gate for one field while every other optional field stays unchecked — inconsistent. If we want fail-fast validation at assembly, it should be a cross-field "validate specs on assemble" pass (its own ticket), not a candidate-specific special case.

